PR #4, opened 2019-02-14
Scripts to do validation and preprocessing on flink using beam

angoenka:
Adding scripts to run tfdv and preprocess on flink.
Please refer to https://beam.apache.org/roadmap/portability/#python-on-flink for python related setup for Beam on Flink.


angoenka:
R: @solma @devidipak

examples/chicago_taxi/execute_on_flink.sh
```diff
@@ -0,0 +1,35 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+set -u
+
+echo Starting pipeline on Flink...
+
+if [ "${JOB_ENDPOINT:-unset}" == "unset" ]; then
+  JOB_ENDPOINT="localhost:8099"
+fi
+
```
devidipak: which steps here start the flink cluster?

examples/chicago_taxi/preprocess_flink.sh
```diff
@@ -0,0 +1,40 @@
...
+# limitations under the License.
+set -u
+
+echo Starting distributed TFT preprocessing...
+
+
+DATA_DIR=$(pwd)/data
+OUTPUT_DIR=$DATA_DIR
+SCHEMA_PATH=$DATA_DIR/flink_tfdv_output/schema.pbtxt
+
```
devidipak: Do we need to specify the Schema_path here ? 

The flow would be for users to run the tfdv_analyse_validate_flink, and the output is stored in these environment variables, right ?

examples/chicago_taxi/tfdv_analyze_and_validate_flink.sh
```diff
@@ -0,0 +1,52 @@
...
+            --input $DATA_DIR/train/data.csv " SCRIPT=tfdv_analyze_and_validate.py $(pwd)/execute_on_flink.sh
+
+EXTRA_ARGS="--for_eval \
+            --validate_stats \
+            --stats_path $TFDV_OUTPUT_PATH/eval_stats.tfrecord \
+            --schema_path $SCHEMA_PATH \
+            --anomalies_path $TFDV_OUTPUT_PATH/anomalies.pbtxt \
+            --save_main_session True \
+            --input $DATA_DIR/eval/data.csv " SCRIPT=tfdv_analyze_and_validate.py $(pwd)/execute_on_flink.sh
+
```
devidipak: so if I understand right, we can do tfdv and tft pre processing on flink runner

This is great! 

Just being nitpicky here, I wonder if we can keep existing scripts, and have arguments passed to it to a)start fliink cluster, and b) pass in the Portable runner with required arguments for that

Also do you have plans to add support for tfma, it would probably be exactly the same steps ?

examples/chicago_taxi/execute_on_flink.sh
angoenka: These scripts are simplified to use only a local flink cluster which has to be started separately. 
Also a separate beam job server has to be started separately based on https://beam.apache.org/roadmap/portability/#python-on-flink 

This script only submits the pipeline to beam job server which submits the job to the configured flink. 
Note: if no flink cluster is running locally then the beam job server will execute the flink job in process.

This script is not intended to be called by the users directly and is used by other scripts.

examples/chicago_taxi/preprocess_flink.sh
angoenka: Yes, SCHEMA_PATH is needed as its passed to the job. 
The user does not have to change this value.

examples/chicago_taxi/tfdv_analyze_and_validate_flink.sh
angoenka: a) These scripts can use the inprocess flink in beam job sever which remove the requirement for a flink cluster. However, user has to start a beam job server. I can automate download and building of beam from git. Please suggest if I should add it in this PR.
b) Using portable runner is take care of in execute_on_flink.sh . From the perspective of example, we don't have to pass any additional arguments for portable runner.

I tried running local tfma but it failed with some module exception. I will try to fix and then can add flink tfma

angoenka:
Thanks for the review @devidipak

examples/chicago_taxi/tfdv_analyze_and_validate_flink.sh
devidipak: SGTM

angoenka:
If everything looks good, shall we go for the merge :)

devidipak:
Will bring it up in our meeting today , and I think we should be able to
merge to TFX  repo

Thanks!
Devi

1025KB:
Hi, Ankur, could you submit the change from internal?

katsiapis:
Should this be closed / reverted now (since the flink example has been published [1][2][3])?

[1] https://github.com/tensorflow/tfx/blob/master/tfx/examples/chicago_taxi/tfdv_analyze_and_validate_flink.sh
[2] https://github.com/tensorflow/tfx/blob/master/tfx/examples/chicago_taxi/preprocess_flink.sh 
[3] https://github.com/tensorflow/tfx/blob/master/tfx/examples/chicago_taxi/process_tfma_flink.sh

Issue #9, opened 2019-03-09
Had to update LOCAL_MODEL_DIR location for chicago_taxi_pipeline example


aaronlelevier:
I was following the [tfx/examples/chicago_taxi_pipeline/README.md](https://github.com/tensorflow/tfx/blob/master/examples/chicago_taxi_pipeline/README.md). Everything worked  pretty flawlessly. I did have to make a change to: 

https://github.com/tensorflow/tfx/blob/master/examples/chicago_taxi/start_model_server_local.sh#L31

I changed `LOCAL_MODEL_DIR` to:

```
LOCAL_MODEL_DIR=$TAXI_DIR/serving_model/taxi_simple/
```

It looks like the code in the `chicago_taxi` dir is different from the `chicago_taxi_pipeline` dir example for where the serving model is stored.

Should I submit a small PR for the change?

 If so, would it be preferred to have a copy of the `start_model_server_local.sh` file in both `chicago_taxi` and `chicago_taxi` with different values for `LOCAL_MODEL_DIR` or it's better to have one `start_model_server_local.sh` file that accepts a `LOCAL_MODEL_DIR` argument?

If I do a PR, I can also include an update to the [README.md](https://github.com/tensorflow/tfx/blob/master/examples/chicago_taxi_pipeline/README.md) to document the change.

**my software version**

Mac OSX
Python 2.7

**p.s.**

I just want to thank you for all of the work that Google has done on TFX. It's amazing for ML to be so accessible and for the ecosystem to come together like this.

Thanks

krazyhaas:
Hi Aaron.  Thanks for the feedback and glad to hear your experience was almost perfect.  A PR would be much appreciated. If you do, my preference would be to have `start_model_server_local.sh`  accept the `LOCAL_MODEL_DIR` argument instead of copying the script.

-K

1025KB:
at the end of  [README.md](https://github.com/tensorflow/tfx/blob/master/examples/chicago_taxi_pipeline/README.md), it has instructions about how to modify the start_model_server_local.sh to fit chicago_taxi_pipeline

aaronlelevier:
So remove the additional `start_model_server_local.sh` file, leave the readme documentation, and just fixup the path where the model is being saved?

1025KB:
the serving and client examples in chicago_taxi dir works for chicago_taxi_pipeline dir excepts some path changes listed in README.md (the current path setup is for chicago_taxi example instead of chicago_taxi_pipeline example)

[...]

PR #10, opened 2019-03-10
update LOCAL_MODEL_DIR location for chicago_taxi_pipeline example

aaronlelevier:
Here is the PR for #9 

The code currently has the Pusher save the model at `$TAXI_DIR/serving_model/taxi_simple`, so this PR updates the saved model path documentation in the `chicago_taxi_pipeline/README.md`. 

It also adds a `chicago_taxi_pipeline/start_model_server_local.sh` file that can be called to serve the model instead of changing the `LOCAL_MODEL_DIR` value in `chicago_taxi/start_model_server_local.sh`.

### DIFF
--- a/examples/chicago_taxi_pipeline/README.md
+++ b/examples/chicago_taxi_pipeline/README.md
@@ -191,7 +191,7 @@ Once the pipeline completes, the model will be copied by the [Pusher](https://gi
 to the directory configured in the example code:
 
 <pre class="devsite-terminal devsite-click-to-copy">
-ls $TAXI_DIR/pipelines/serving_model/taxi_simple
+ls $TAXI_DIR/serving_model/taxi_simple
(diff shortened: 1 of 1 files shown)


aaronlelevier:
I signed it!

aaronlelevier:
I updated the PR for my Github user account with the signed CLA.

I'm a little confused on the feedback though. Do you want me to create a new `start_model_server_local.sh` and `classify_local.sh` under `examples/chicago_taxi_pipeline/`? That's what one of the last comments said.

I looked over the `examples/chicago_taxi_pipeline/README.md` again and the saved model path currently looks correct in my PR if that's the only desired change.

1025KB:
current PR with only examples/chicago_taxi_pipeline/README.md change looks good to me

PR #30, opened 2019-04-05
Lock in version of `kfp`

MattMorgis:
My team and I hit a snag this week when trying to compile Kubeflow Pipelines:

```
File "/tfx/examples/chicago_taxi_pipeline/taxi_pipeline_kubeflow.py", line 187, in <module>
    pipeline = KubeflowRunner().run(_create_pipeline())

File "/tfx/orchestration/kubeflow/runner.py", line 122, in run
    _construct_pipeline)

TypeError: add_pipeline() takes 2 positional arguments but 3 were given
```

Thanks to commit [`eb63dae`](https://github.com/tensorflow/tfx/commit/eb63daed199afee8a88fca44c2e8e0f738adf44b) this week which added a test for this, we were able to debug easier.

It turns out the issue was that we had version `0.1.14` of the Kubeflow Pipelines SDK installed. The [docs](https://www.kubeflow.org/docs/pipelines/sdk/install-sdk/) recommended this, and the docs and the [Release Page](https://github.com/kubeflow/pipelines/releases) do not install from PyPi. 

This feels like it was intentional, but only version published to PyPi is `0.1.11`, the last compatible version with tfx. I've tried `0.1.12` all the way to the most recent `0.1.15` and continue to get the same error as above. 

Additionally, the check submitted by this PR doesn't even catch this. 

When running `pip show kfp` from the non-PyPi install, I get:
```
Name: kfp
Version: 0.1
Summary: KubeFlow Pipelines SDK
Home-page: UNKNOWN
Author: google
Author-email: None
License: UNKNOWN
Location: /usr/local/lib/python3.7/site-packages
Requires: PyYAML, kubernetes, requests-toolbelt, certifi, google-cloud-storage, google-auth, cryptography, python-dateutil, six, PyJWT, urllib3
Required-by:
```

And when running after installing from `setup.py`:
```
Name: kfp
Version: 0.1.11
Summary: KubeFlow Pipelines SDK
Home-page: UNKNOWN
Author: google
Author-email: None
License: UNKNOWN
Location:  /usr/local/lib/python3.7/site-packages
Requires: PyJWT, google-cloud-storage, PyYAML, certifi, six, python-dateutil, kubernetes, requests-toolbelt, cryptography, urllib3, google-auth
```

I'm not sure if I should follow an issue in [Kubeflow/pipelines](https://github.com/kubeflow/pipelines) and ask if they can tag with the patch, or if these will never be published to PyPi and this is a non-issue. 

I'm unsure what the proper fix should be, but wanted to submit to start the conversation and document for anyone else who may run into this.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -89,7 +89,7 @@ def _make_required_test_packages():
   return [
       'apache-airflow>=1.10,<2',
       'docker>=3.7,<4',
-      'kfp>=0.1,<1',
+      'kfp>=0.1,<=0.1.11',
       'tensorflow>=1.13,<2',
   ]
 


MattMorgis:
I signed it

Ark-kun:
Thanks for the fix.

The runner is failing, because it was using undocumented hack.

The KFP clearly says that the pipeline function must be decorated with `@pipeline` decorator. It's even mentioned in the `compile` function documentation: https://github.com/kubeflow/pipelines/blob/086d4763d9f163acdeb423bc2bc49ab470442a92/sdk/python/kfp/compiler/compiler.py#L636

Running `compile` without having the function decorated
https://github.com/tensorflow/tfx/blob/d9f87b9b03ab8bb98653b8a96d5bf952bc22b2b7/tfx/orchestration/kubeflow/runner.py#L125 will raise error: https://github.com/kubeflow/pipelines/blob/086d4763d9f163acdeb423bc2bc49ab470442a92/sdk/python/kfp/compiler/compiler.py#L578 `'Please use a function with @dsl.pipeline decorator.'`.

This documented behavior was circumvented by a hack that used an undocumented function that should be a part of internal compiler API. Ning has recently changed this internal function, which triggered TFX failure.

I'll send a PR fixing TFX code fixing this issue and removing the use of undocumented API.

Ark-kun:
Added the PR to fix the TFX side: https://github.com/tensorflow/tfx/pull/39
It adds the `@pipeline` decorator to the pipeline function just like the documentation says.

PR #39, opened 2019-04-11
Orchestration/Kubeflow - Removed the use of undocumented KFP compiler API

Ark-kun:
See https://github.com/tensorflow/tfx/pull/30#issuecomment-482289948

The KFP documentation clearly says that the pipeline function must be decorated with `@pipeline` decorator. It's even mentioned in the `compile` function documentation: https://github.com/kubeflow/pipelines/blob/086d4763d9f163acdeb423bc2bc49ab470442a92/sdk/python/kfp/compiler/compiler.py#L636

Running `Compiler().compile(...)` without having the function decorated will raise error: https://github.com/kubeflow/pipelines/blob/086d4763d9f163acdeb423bc2bc49ab470442a92/sdk/python/kfp/compiler/compiler.py#L578 `'Please use a function with @dsl.pipeline decorator.'`.

For some reason this documented behavior was circumvented by a hack that used an undocumented function that should be a part of internal compiler API.

This has triggered the https://github.com/tensorflow/tfx/pull/30 issue when the internal compiler API was changed by the KFP team.

This PR fixes this issue.

### DIFF
--- a/tfx/orchestration/kubeflow/runner.py
+++ b/tfx/orchestration/kubeflow/runner.py
@@ -95,8 +95,8 @@ class KubeflowRunner(tfx_runner.TfxRunner):
     """
 
     @dsl.pipeline(
-      pipeline.pipeline_args['pipeline_name'],
-      pipeline.pipeline_args.get('description', '')
+      name=pipeline.pipeline_args['pipeline_name'],
+      description=pipeline.pipeline_args.get('description', '')
     )
     def _construct_pipeline():
       """Constructs a Kubeflow pipeline.
--- a/tfx/orchestration/kubeflow/runner.py
+++ b/tfx/orchestration/kubeflow/runner.py
@@ -94,6 +94,10 @@ class KubeflowRunner(tfx_runner.TfxRunner):
         pipeline.
     """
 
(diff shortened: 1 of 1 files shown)


Ark-kun:
Gentle ping

Ark-kun:
Ping.

tfx/orchestration/kubeflow/runner.py
```diff
@@ -94,6 +94,10 @@ def run(self, pipeline):
         pipeline.
     """
 
+    @dsl.pipeline(
+      pipeline.pipeline_args['pipeline_name'],
```
neuromage: Maybe add parameter names here?

```
@dsl.pipeline(
  name=...
  description=...
```

Ark-kun: Done.

Ark-kun:
JFYI: The decorator is not required by the KFP compiler since v0.1.17 (last week release).

Also:

```
@decorator(args)
def func(...)
    ...
```
is syntactic sugar for
```
def func(...):
    ...

func = decorator(args)(func)
```

Issue #47, opened 2019-04-18
ImportError: No module named tfx.examples.chicago_taxi.trainer on Dataflow


mwalenia:
I'm trying to run the Chicago Taxi Example on Google Dataflow and I ran into a problem with imports. After setting `PYTHONPATH` env in order to make the python files run on my environment, the pipelines fail with 

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/dataflow_worker/batchworker.py", line 766, in run
  self._load_main_session(self.local_staging_directory)
  File "/usr/local/lib/python2.7/dist-packages/dataflow_worker/batchworker.py", line 482, in _load_main_session
 pickler.load_session(session_file)
 File "/usr/local/lib/python2.7/dist-packages/apache_beam/internal/pickler.py", line 269, in load_session
 return dill.load_session(file_path)
File "/usr/local/lib/python2.7/dist-packages/dill/_dill.py", line 410, in load_session
   module = unpickler.load()
File "/usr/lib/python2.7/pickle.py", line 864, in load
   dispatch[key](self)
 File "/usr/lib/python2.7/pickle.py", line 1139, in load_reduce
value = func(*args)
 File "/usr/local/lib/python2.7/dist-packages/dill/_dill.py", line 828, in _import_module
return getattr(__import__(module, None, None, [obj]), obj)
ImportError: No module named tfx.examples.chicago_taxi.trainer
```

I don't know how to remedy this - is it possible to set an environmental variable in a Dataflow pipeline?

krazyhaas:
Quick update: I've been able to recreate the problem but haven't yet found the fix.

krazyhaas:
Thanks for the PR!  I’ll check it out tomorrow.

ChethanUK:
@mwalenia Line: https://github.com/tensorflow/tfx/blob/942358f5b24dea49e8bde8691507c8cbc508b427/tfx/examples/chicago_taxi/trainer/task.py#L26 is giving 
```
Traceback (most recent call last):
  File "trainer/task.py", line 25, in <module>
    from tfx.examples.chicago_taxi.trainer import model
ImportError: No module named tfx.examples.chicago_taxi.trainer
```

PR #48, opened 2019-04-18
Fix Chicago Taxi Example not working on Dataflow

mwalenia:
I ran into an [issue](https://github.com/tensorflow/tfx/issues/47) when trying to run Chicago Taxi Example on Dataflow. From my investigation it seems that the code package uploaded to GS didn't reflect directory structure of the repository - importing from `tfx.examples.chicago_taxi.trainer` made no sense, since the `tar.gz` archive looks like this: 

```
├── PKG-INFO
├── README.md
├── setup.cfg
├── setup.py
├── tfx_chicago_taxi.egg-info
│   ├── PKG-INFO
│   ├── SOURCES.txt
│   ├── dependency_links.txt
│   ├── requires.txt
│   └── top_level.txt
└── trainer
    ├── __init__.py
    ├── model.py
    ├── task.py
    └── taxi.py
```

After changing the import to relative, my Dataflow jobs started working.
WDYT?

### DIFF
--- a/tfx/examples/chicago_taxi/preprocess.py
+++ b/tfx/examples/chicago_taxi/preprocess.py
@@ -28,7 +28,7 @@ import tensorflow_transform.beam as tft_beam
 from tensorflow_transform.coders import example_proto_coder
 from tensorflow_transform.tf_metadata import dataset_metadata
 from tensorflow_transform.tf_metadata import dataset_schema
-from tfx.examples.chicago_taxi.trainer import taxi
+from trainer import taxi
 
 
 def _fill_in_missing(x):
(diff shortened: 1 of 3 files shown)


mwalenia:
I signed it!

mwalenia:
@krazyhaas Hi, when can I expect an update about this?

mwalenia:
@zhitaoli thanks for merging my PR!
However when I tried to run the updated version in a dataflow context, I encountered the same error  `ImportError: No module named tfx.examples.chicago_taxi.trainer`

Is it possible that the try-except doesn't work properly on Dataflow?

PR #49, opened 2019-04-19
Add Parquet example gen component

tamera-lanham:
This PR adds a new `ParquetExampleGen` component to ingest data from Parquet files; most of the code is minimally adapted from `CsvExampleGen`. 

The component works as intended and all tests are passing, but some refactoring might be advisable because a lot of the code is duplicated between these two components - in particular,`driver.py` and `component.py` are taken wholesale from the code for `CsvExampleGen`. If changes should be made to avoid the duplication, please let me know! This PR can be a work in progress, I would appreciate any guidance the maintainers have to offer.


1025KB:
Thanks for the contribution! We have some internal refactoring about example gen, which should make custom example gen easier. I will get back to you PR soon!

1025KB:
Hi, Sorry for the long waiting, we refactored the example gen, now you only need the executor to support parquet file format, here is the instruction for [custom example gen](https://github.com/tensorflow/tfx/blob/master/docs/guide/examplegen.md#custom-examplegen)

we are also working on [type hint](https://github.com/tensorflow/tfx/issues/95), after that issue is fixed, we can start merging PRs

PR #106, opened 2019-05-10
Fix requirements error with scikit-learn version

Maestro157:
As the latest released version of scikit-learn, version 0.21.0, only supports python>=3.5, the previous version 0.20.3 has to be explicitly installed, so that tfx can be installed on python==2.7


Maestro157:
I signed it!

1025KB:
Hi, we are working on the 3.5 releasing soon!

1025KB:
FYI, https://github.com/scikit-learn/scikit-learn/issues/13860

PR #178, opened 2019-05-25
fix typo

dakl:
iff => if

### DIFF
--- a/tfx/components/transform/component.py
+++ b/tfx/components/transform/component.py
@@ -42,7 +42,7 @@ class Transform(base_component.BaseComponent):
       artifact.
     module_file: The file path to a python module file, from which the
       'preprocessing_fn' function will be loaded.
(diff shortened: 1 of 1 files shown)


dakl:
I signed it!

tfx/components/transform/component.py
```diff
@@ -42,7 +42,7 @@ class Transform(base_component.BaseComponent):
       artifact.
     module_file: The file path to a python module file, from which the
       'preprocessing_fn' function will be loaded.
-    name: Optional unique name. Necessary iff multiple transform components are
```
zhitaoli: By "iff" I meant for "if and only if" here. Does that make sense to you?

If you think we should just use "if" I'm happy to accept this PR.

1025KB:
close for now, reopen if needed

PR #202, opened 2019-06-05
Added Pipeline Property to mount a PV into all components

rummens:
# What
I have added a new pipeline property called _pipeline_pv_mount_ to allow mounting of a PV into all components of the Pipeline. 

pipeline_pv_mount must be a dict with the following keys: pvc_name, volume_name, volume_mount_path.
 volume_name defaults to 'pipeline-shared-volume' and volume_mount_path to '/pipeline_shared_volume'. 'pvc_name' will raise an ValueException if not provided. 

# Why
When running KubeFlow outside of GCP sharing data between components via buckets is not possible. Therefore one can now mount a PV to share data between the components. It is encouraged to use a PV of type ReadWriteMany.

# Additional Help
I am not familiar with the documentation and test process of TFX, so please advice me on additional changes.

# Example PV 
The following PV is an example which can be used to share data. Adapt the storageClassName to your plattform and needs.

```
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: pipeline-shared-data
  labels:
    app: kubeflow-pipelines
spec:
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      storage: 2Gi
  storageClassName: azurefile

```

### DIFF
--- a/tfx/orchestration/kubeflow/base_component.py
+++ b/tfx/orchestration/kubeflow/base_component.py
@@ -55,7 +55,7 @@ class PipelineProperties(object):
                  tfx_image: Text = None,
                  pipeline_pv_mount=None):
         """pipeline_pv_mount must be a dict with the following keys: pvc_name, volume_name, volume_mount_path.
-        volume_name defaults to 'pipeline_shared_volume' and volume_mount_path to '/pipeline_shared_volume'.
+        volume_name defaults to 'pipeline-shared-volume' and volume_mount_path to '/pipeline_shared_volume'.
         It is encouraged to use a PV of type ReadWriteMany."""
 
         self.exec_properties = collections.OrderedDict([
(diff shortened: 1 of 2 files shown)


neuromage:
Thanks for your PR.

For my understanding, can you tell me what is the use case for mounting shared volumes here? Are the input/output files stored here? Note that if you store it in a PV, then metadata tracking loses value since the URIs are transient and will no longer be meaningful after the pipeline completes.

I'd prefer not to expose KFP options this way using PipelineProperties to the user as it's not scalable. We're looking into a more general method for exposing runtime configuration like this and should have something out soon. In the meantime though, understanding your use-case for shared volumes in a TFX pipeline would be very beneficial in that regard.

/cc @ruoyu90 
/cc @ucdmkt 
/cc @hongye-sun
/cc @zhitaoli

rummens:
I need it exactly for this, storing the inputs/output files. The PV is persistent until you delete it, we use azure file under the hood to ensure this. This might not be optimal but is the only possibility I see to share the data without relying on vendor specific object storage. 
Only alternative is to deploy minio to miror S3 but I am not sure if this is supported?

What do you have in mind to have the config more general?

rummens:
I have the idea from the Kubeflow Pipelines TFX example: https://github.com/kubeflow/pipelines/tree/master/samples/tfx

rummens:
More general fix created by #208. This is a better approach than mine, will close PR. Thanks @tfx-copybara.

rummens:
For documentation purposes this is how to add a PV with #208:

```
from tfx.orchestration.kubeflow.runner import KubeflowRunner, KubeflowRunnerConfig
from kfp.onprem import mount_pvc

runner_config = KubeflowRunnerConfig([mount_pvc(pvc_name="pipeline-shared-data",
                                                    volume_mount_path=code2flow.pvc_mount)])
_ = KubeflowRunner(config=runner_config).run(pipeline_object)
```

PR #365, opened 2019-07-19
[Docs] Update workshop.md

JTunis:
1. Must change into the directory containing the tfx repo's `.git` or one of its subdirectories before checking out the `r0.13` branch
2. [Optional] Since `setup_demo.sh` installs tensorflow + tfx, I thought it was redundant to also include those installations here


1025KB:
Thanks, we updated the workshop.md recently

Issue #410, opened 2019-08-01
Ability to name dataflow jobs


htahir1:
Is there a way / or a plan to support custom names for job ids of dataflow jobs? right now, the components spin up something like this:

![image](https://user-images.githubusercontent.com/4792082/62274344-fd78c300-b43f-11e9-8cb5-961cdc6e8691.png)

Its very hard to distinguish between pipeline runs and its hard also to collaborate and see your runs vs other people in the company.

My suggestion would be to at least by default support adding the component name to the job id. This way at least one can see at a glance if it was statisticsgen, schemagen etc. Thanks!

charlesccychen:
Thanks, I will look into adding this functionality.

benkoller:
I took the liberty of adding a minimally-invasion attempt at mitigating this for both the orchestrator as well as the executor pipelines.

jeongukjae:
This is a very old issue, but I leave a comment for other people.

You can name the DataFlow job using `job_name` parameter. (https://cloud.google.com/dataflow/docs/guides/specifying-exec-params#python_4)

Example)

```python
    pipeline.Pipeline(
        ...
        beam_pipeline_args=[
            ...
            "--runner=DataflowRunner",
            "--job_name=job-name-for-dataflow"
        ],
    )
```

AlirezaSadeghi:
> This is a very old issue, but I leave a comment for other people.
> 
> DataFlow job can be named by `job_name` parameter. ([cloud.google.com/dataflow/docs/guides/specifying-exec-params#python_4](https://cloud.google.com/dataflow/docs/guides/specifying-exec-params#python_4))
> 
> Example)
> 
> ```python
>     pipeline.Pipeline(
>         ...
>         beam_pipeline_args=[
>             ...
>             "--runner=DataflowRunner",
>             "--job_name=job-name-for-dataflow"
>         ],
>     )
> ```

This won't work as probably not many would want all their dataflow jobs to be named the same.

You might have statisticsgen, schemagen, transform, evaluator and ... run dataflow jobs in the background and you don't want to name them the same.

Only remedy I see is to allow per component configuration of the underlying dataflow job, through custom_configs or something like that. 

This is not only an issue for name, it's an issue for the job config (number of workers/machine types), labels and ... and should be solved for that IMO.

singhniraj08:
@htahir1,

[...]

PR #466, opened 2019-08-12
build job names dynamically of component, pipelinename, timestamp

benkoller:
This is a quick attempt to mitigate #410 in a minimally-invasive fashion.


benkoller:
@googlebot I signed it!

tfx/orchestration/beam/beam_dag_runner.py
```diff
@@ -93,6 +106,22 @@ def run(self, tfx_pipeline: pipeline.Pipeline) -> None:
...
+      '{pipeline_name}-{ts}'.format(
+        pipeline_name=tfx_pipeline.pipeline_info.pipeline_name,
+        ts=int(datetime.datetime.timestamp(datetime.datetime.now()))
+      )
+    )
+    # Exploit the last-in priority of --job_name for beam
+    if self._beam_orchestrator_args:
+      self._beam_orchestrator_args.append('--job_name={}'.format(job_name))
+    else:
+      self._beam_orchestrator_args = ['--job_name={}'.format(job_name)]
```
1025KB: the beam orchestrator is for entire pipeline instead of a single component

benkoller: Hi @1025KB , maybe I have a misunderstanding then and you can help me by clarifying. I was under the assumption from the comments in the code that `self._beam_orchestrator_args` are responsible for the orchestrator pipeline args and that `component_launcher.ComponentLauncher()` takes `additional_pipeline_args`, which in turn contain a list of `beam_pipeline_args`. This drove me to these design decisions.

Please also note that the reviewed code contained a bug under which the actual pipeline names for executors could be wrong due to a race-condition-like flaw in my own naming logic. According to my tests its fixed now.

benkoller:
Cool that 0.14. is out now, is there a way I can assist in putting this onto the roadmap for 0.14.1 / 0.15.0?

1025KB:
Hi, if you are trying to run tfx on dataflow, could you try kubeflow_dag_runner? which I think already handles those problems

benkoller:
Hi @neuromage thanks for taking a look.

Dataflow creates pipeline names by itself, in the format `beam-<USERNAME>-<timestamp`. This creates opaque pipeline names for both the orchestrator as well as the components.

By creating pipeline names based on a more descriptive pattern in accordance to the tfx `pipeline_name` makes individual runs identifiable. Especially in an environment of multiple parallel experiments and a CI/CD approach this is import for us to correlate code, config and experiment.

Keeping our use case aside it seems counter-intuitive to specify a `pipeline_name` and end up with Dataflow pipelines that can't easily be correlated via name. This fixes this.

In a more elaborate solution the beam arguments for orchestrator and components would be passed not as a list of strings but as a dictionary that is then deconstructed into the Dataflow/Beam command line parameters downstream. My fix is really just a simple fix in this comparison.

neuromage:
Thanks @benkoller. I think ideally, we'd like a solution that works across orchestrators, and not just for the BeamDagRunner. Given this, I suggest we close this PR and wait for the more general solution. Does that work for you?

benkoller:
@neuromage if a generalised solution is possible that would definitely work for me. Out of curiosity, at what point in the code would you handle the job namings? Or would you still handle namings in the orchestrator runners, but via a util?

neuromage:
I haven't thought about this enough, but off the top of my head, for individual Dataflow jobs, I would most likely update this using a combination of `beam_pipeline_args` and executor `context`:
https://github.com/tensorflow/tfx/blob/d651dfda08278dd46a32b56931c6a61d7ed2e3c1/tfx/components/base/base_executor.py#L34


Thanks again for the PR! I'm going to go ahead and close this one for now.

PR #477, opened 2019-08-14
Pass exec_properties to BaseDriver's _prepare_output_artifacts

martin-laurent:
It's currently very difficult to override the `_prepare_output_artifacts` function in the `BaseDriver`.
This function currently receive only basic pipeline and component information as well as the `output_dict`. 
That means you cannot have external parameters in your custom logic.

This PR passes the `exec_properties` dictionary as a parameter, in a similar way to what is done in `resolve_input_artifacts`.
It also rename `_prepare_output_artifacts`  to `prepare_output_artifacts` to make it "public" according to Python standard

### DIFF
--- a/tfx/components/base/base_driver.py
+++ b/tfx/components/base/base_driver.py
@@ -165,13 +165,13 @@ class BaseDriver(object):
     """
     return exec_properties
 
-  def _prepare_output_artifacts(
+  def prepare_output_artifacts(
       self,
       output_dict: Dict[Text, types.Channel],
       execution_id: int,
       pipeline_info: data_types.PipelineInfo,
       component_info: data_types.ComponentInfo,
-      exec_properties: Dict[Text, Any],
+      exec_properties: Dict[Text, Any],  # pylint: disable=unused-argument
   ) -> Dict[Text, List[types.Artifact]]:
     """Prepare output artifacts by assigning uris to each artifact.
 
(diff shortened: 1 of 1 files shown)


1025KB:
Hi, based on our current design, output artifact is decided by component definition, could you explain your use case for overwriting this function? thanks!

martin-laurent:
Hi. In our company, when we do data for production, we have to abide by a set of rules as to where we can store things (like models, training data and results). It also allows us to use a deterministic method to compute the path to our data (which is not the case with the execution id being part of the path today). This is all because of GDPR and more generally because we don't want to store everything in the same S3 or GCS bucket. The current implementation prevent us from implementing such a logic. To do it we need to override the output uri computation by our own logic that uses `exec_properties`. I'm sure other users may want a more flexible way to compute the output uris of their components.

martin-laurent:
Hey @ruoyu90, any update on this? Do you think it could make it for next release?

andrewsmartin:
Hi @ruoyu90, what's the status of this PR? Is it possible to get this merged? (Looks like it needs a rebase).

andrewsmartin:
Hey @ruoyu90, it looks like you've approved this but it hasn't been merged yet. Are we able to get this merged (after resolving conflicts)?

rclough:
Commenting to make this not stale.

1025KB:
Just FYI, it's already added in one of the internal [change](https://github.com/tensorflow/tfx/blob/master/tfx/dsl/components/base/base_driver.py#L201) 

Thanks!

PR #652, opened 2019-09-19
ModelValidator to compare MetricValues to thresholds

martin-laurent:
Modify the ModelValidator to take an extra dictionary from metric key to metirc value.
In `_pass_threshold` in `ModelValidator`'s `Executor`, the metrics from EvalResult are then converted to their corresponding `metrics_for_slice_pb2.MetricValue` protobuf equivalent and compared against the corresponding threshold value.

Happy to talk about the approach (for threshold_config: dictionary or protobuf message) and also of the way forward to compare against the blessed model's results

### DIFF
--- a/tfx/components/model_validator/component.py
+++ b/tfx/components/model_validator/component.py
@@ -17,7 +17,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-from typing import Optional, Text
+from typing import Optional, Text, Dict, Any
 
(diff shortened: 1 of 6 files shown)


1025KB:
Hi, Martin, thanks for the cl, we are working on a bigger refactor now for merging evaluator and modelvalidator, internally we have a proto config for define threshold but this one doesn't available in oss, the oss version of threshold config are wip, let's disucss in the issue thread

PR #854, opened 2019-10-31
schemagen.md Fixed typos. Enhanced readability [master]

foohasmallya:


docs/guide/schemagen.md
```diff
@@ -43,33 +45,30 @@ feature {
...
 
 ## Using the SchemaGen Component
 
-A SchemaGen pipeline component is typically very easy to deploy and requires little
+A SchemaGen component is typically very easy to deploy and requires little
 customization. Typical code looks like this:
 
 ```python
-from tfx import components
+from tfx.components import SchemaGen
```
numerology: The reason for current way of writing this can be found at [Google python style guide](http://google.github.io/styleguide/pyguide.html#224-decision)

For most cases we import modules, but not classes.

foohasmallya: Thanks for correcting me, @numerology. Will make changes to the RC PR of this doc too. I used "from tfx.components import <component_name> in my updates to the StatisticsGen docs. However, I missed out here.

PR #883, opened 2019-11-02
ImporterNode: Propagate imported result to output dict

jinnovation:
Current implementation of ImporterNode propagates the specified URI to ML Metadata, but does not expose said URI as part of `.output`.

This PR patches `ImporterDriver` to propagate the imported URI accordingly. This logic could certainly live in some newly-created `ImporterExecutor` that we use in lieu of `EmptyExecutor`, which would possibly be more in line with the separation-of-responsibilities between Driver and Executor. Reviewers, let me know if you'd rather go down that route and I can update accordingly.

### DIFF
--- a/tfx/components/common_nodes/importer_node.py
+++ b/tfx/components/common_nodes/importer_node.py
@@ -82,8 +82,8 @@ class ImporterDriver(base_driver.BaseDriver):
                 reimport=exec_properties[REIMPORT_OPTION_KEY])
     }
 
-    # output_dict[IMPORT_RESULT_KEY] = channel_utils.as_channel(
-    #   output_artifacts[IMPORT_RESULT_KEY])
+    output_dict[IMPORT_RESULT_KEY] = channel_utils.as_channel(
+      output_artifacts[IMPORT_RESULT_KEY])
 
     return data_types.ExecutionDecision(
         input_dict={},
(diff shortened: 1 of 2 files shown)


ruoyu90:
Hi @jinnovation , output future is already [populated](https://github.com/tensorflow/tfx/blob/master/tfx/components/common_nodes/importer_node.py#L137).  Also here is an [example](https://github.com/tensorflow/tfx/blob/master/tfx/examples/chicago_taxi_pipeline/taxi_pipeline_importer.py) using ImporterNode. Let me know if this answer your questions :)

jinnovation:
@ruoyu90: I'll have to confirm this, but I recall running with `InteractiveContext` recently and not having the result populate as expected. Either way, let's go ahead and close this for now. If it turns out to be either some other issue, I'll open a separate PR.

ruoyu90:
@jinnovation Looking at this again I think I know where the issue is: We have [special handling](https://github.com/tensorflow/tfx/blob/92832988691c82c6b142a6251385c46e86c38991/tfx/components/base/base_driver.py#L130) for artifact resolution under `interactiveContext`. Basically it bypasses artifact resolution and directly use the exact `output` dict of the node referencing it. Your fix will be the easiest way to make it work. Thanks!

PR #886, opened 2019-11-04
Expose exec_properties to BaseDriver._prepare_output_artifacts

jinnovation:
This paves the way to having output preparation be influenced by Component-level execution properties—a precedent already set in `BaseDriver.resolve_input_artifacts`.

As an aside, is there any opposition to exposing `_prepare_output_artifacts` as "public"?


jinnovation:
Looks like this PR is redundant with #477.

PR #991, opened 2019-11-26
Fix #990, correctly implement ComponentSpec equality check

### LINKED ISSUE #990: ComponentSpec equality is broken
https://github.com/tensorflow/tfx/pull/709 added support for comparing `ComponentSpec` instances via `__eq__`, but the implementation is broken due to a flawed `isinstance()` check. Here's an example to reproduce:

```
>>> from tfx.types.standard_component_specs import EvaluatorSpec
>>> import tensorflow as tf
>>> from tfx.components.schema_gen import component
>>> from tfx.types import channel_utils, standard_artifacts
>>> schema_gen = component.SchemaGen(statistics=channel_utils.as_channel([standard_artifacts.ExampleStatistics(split='train')]))
>>> schema_gen.spec == schema_gen.spec
False
```

This is because of the [isinstance check here](https://github.com/tensorflow/tfx/pull/709/files#diff-5255d3b9d52612c19f057d278f08c905R91) on the spec classes, since class objects are never instances of other class objects (but rather metaclasses). I think this can be corrected by just ensuring the two classes are equal.

Happy to PR this fix.

andrewsmartin:
This PR also adds an extra assert in `component_spec_test.py`.

### DIFF
--- a/tfx/types/component_spec.py
+++ b/tfx/types/component_spec.py
@@ -88,7 +88,7 @@ class ComponentSpec(with_metaclass(abc.ABCMeta, json_utils.Jsonable)):
     self._parse_parameters()
 
   def __eq__(self, other):
-    return (isinstance(other.__class__, self.__class__) and
+    return (other.__class__ == self.__class__ and
             self.to_json_dict() == other.to_json_dict())
 
   def _validate_spec(self):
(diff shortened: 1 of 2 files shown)


liyzcj:
Any one respect this question?

PR #1012, opened 2019-12-05
Fix in rest_simple tutorial notebook

hemanth171:
In the rest_simple.ipynb notebook (https://www.tensorflow.org/tfx/tutorials/serving/rest_simple) there is a piece of code which show the comparison between the prediction and the actual one. But the comparison is done to same prediction vs prediction. But actually it should be between prediction vs test_image.

This is the wrong code which is show in the tutorial:
show(0, 'The model thought this was a {} (class {}), and it was actually a {} (class {})'.format(
**class_names[np.argmax(predictions[0])]**, test_labels[0], **class_names[np.argmax(predictions[0])]**, test_labels[0]))

But the code should be like below:
show(0, 'The model thought this was a {} (class {}), and it was actually a {} (class {})'.format(
**class_names[np.argmax(predictions[0])]**, test_labels[0], **class_names[test_images[0]]**, test_labels[0]))

show(i, 'The model thought this was a {} (class {}), and it was actually a {} (class {})'.format(
    **class_names[np.argmax(predictions[i])]**, test_labels[i], **class_names[test_images[i]]**, test_labels[i]))

### DIFF
--- a/docs/tutorials/serving/rest_simple.ipynb
+++ b/docs/tutorials/serving/rest_simple.ipynb
@@ -1,870 +1,874 @@
 {
-  "cells": [
-    {
-      "cell_type": "markdown",
-      "metadata": {
-        "colab_type": "text",
-        "id": "MhoQ0WE77laV"
-      },
-      "source": [
-        "##### Copyright 2019 The TensorFlow Authors."
-      ]
-    },
-    {
-      "cell_type": "code",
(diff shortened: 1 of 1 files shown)


hemanth171:
@google-admin @zhitaoli I have fixed the mistake which i made in my previous PR. Now all checks have been passed. Could you please review this PR.

PR #1054, opened 2019-12-21
Fix the issue about tutorial notebooks won't run in Colab.

AseiSugiyama:
I fixed the issue of TFDV tutorial about that it raises an error in Colab environment.

In Colab environment, a user has to restart runtime after installing TFDV and Colab will show the "RESTART RUNTIME" button without pip -q option.
With pip -q option, however, running code cell contains pip command raises FileNotFoundError and Colab will not show the "RESTART RUNTIME" button. I think a user can easily miss the requirement to restart runtime (like me).

image

In this PR, I removed pip -q option and divide the code cell to show "RESTART RUNTIME" button at the bottom of the logs.

### DIFF
--- a/docs/tutorials/data_validation/tfdv_basic.ipynb
+++ b/docs/tutorials/data_validation/tfdv_basic.ipynb
@@ -180,7 +180,19 @@
         "colab": {}
       },
       "source": [
-        "!pip install tensorflow_data_validation\n",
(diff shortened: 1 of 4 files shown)


AseiSugiyama:
I found the same issue with tutorial notebooks. I made some additional commits to fix them.

Issue #1055, opened 2019-12-22
taxi_pipeline_beam can't run well due to "base_model" in trainer_fn_args


rmgogogo:
In taxi_utils.py, If change warm_start_from = None, it can run.
Otherwise, it will say KeyError: 'base_model'.

In Trainer component document, the sample trainer_fun is using base_models but in codes it uses base_model.

This part can be cleaned. Guess it's due to temp uncompatibility changes.

My env:
tfx                        0.15.0
tfx-bsl                    0.15.3
tf                          1.15

ucdmkt:
@rmgogogo, the examples at master head is not necessarily guaranteed to be compatible with the released TFX (the tfx 0.15.0). You could try `git clone --branch 0.15.0 https://github.com/tensorflow/tfx.git` and exercise pipelines under `./examples/`. 

@ruoyu90 can confirm incompatibly between head and released versions.

ucdmkt:
\cc yinghsienwu@

ruoyu90:
Thanks @ucdmkt for the explanation. @rmgogogo can you try to follow the instruction from @ucdmkt to see whether it works?

gowthamkpr:
@rmgogogo Did it work or Are you still facing the same issue?

rmothukuru:
Automatically closing due to lack of recent activity. Please update the issue when new information becomes available, and we will reopen the issue. Thanks!

PR #1056, opened 2019-12-22
defaults absent value as None, otherwise basic sample can't work

rmgogogo:
https://github.com/tensorflow/tfx/issues/1055


rmgogogo:
@googlebot I fixed it.

PR #1286, opened 2020-02-14
fix _fill_in_missing bug when pass Tensor

fsx950223:


chongkong:
It seems the notebook file is working without error (i.e. parameter was SparseTensor not Tensor) do you have more context for this PR?

fsx950223:
> It seems the notebook file is working without error (i.e. parameter was SparseTensor not Tensor) do you have more context for this PR?

I got some Tensors last time, but they're all SparseTensor now. I may forget to restart runtime last time. Sorry for disturbed.

PR #1388, opened 2020-03-03
Support BOOLEAN row values by casting to integer

joshtemple:
Boolean types can be coerced to an Int64List according to docs [here](https://www.tensorflow.org/tutorials/load_data/tfrecord#data_types_for_tfexample).


joshtemple:
@googlebot I signed it!

PR #1518, opened 2020-03-21
Fix datatype of the BigQuery query

AseiSugiyama:
This PR fixes the problem that [tfx/template.ipynb](https://github.com/tensorflow/tfx/blob/master/docs/tutorials/tfx/template.ipynb) fails at step 7 "Try BigQueryExampleGen".

Using `CsvExampleGen`, the data type of `pickup_sensus_tract` is nullable string. With `BigQueryExampleGen`, the datatype of `pickup_sensus_tract` should be nullable string too, however, it is INT with the original query.

This causes Invalid argument error at Step 7 of [tfx/template.ipynb](https://github.com/tensorflow/tfx/blob/master/docs/tutorials/tfx/template.ipynb). Full error message is below;

```
2020-03-20 12:00:48.705697: W tensorflow/core/framework/op_kernel.cc:1655] OP_REQUIRES failed at example_parsing_ops.cc:91 : Invalid argument: Name: <unknown>, Key: pickup_census_tract, Index: 90.  Data types don't match. Expected type: string, Actual type: int64
```

I made the query returns string value using safe cast function.

### DIFF
--- a/tfx/experimental/templates/taxi/pipeline/configs.py
+++ b/tfx/experimental/templates/taxi/pipeline/configs.py
@@ -79,7 +79,7 @@ _query_sample_rate = 0.0001  # Generate a 0.01% random sample.
 #           dropoff_latitude,
 #           dropoff_longitude,
 #           trip_miles,
-#           pickup_census_tract,
+#           SAFE_CAST(pickup_census_tract AS STRING) AS pickup_census_tract,
 #           dropoff_census_tract,
 #           payment_type,
 #           company,


jiyongjung0:
Thank you so much for the contribution!

We reviewed your proposal but concluded that the field should be an "INT" type. The CsvExampleGen marks the field as *nullable string* because there is no data which has a value for that field. So we inserted a data which has a value in `pickup_sensus_tract` field in #1272 . This change is integrated in TFX 0.21.1 or later.

(I initially tried to delete the field, and changed my mind to cast the field(as you did!), and finally fixed the issue by inserting a record. That is the reason the commit message and PR subject is inconsistent...)

AseiSugiyama:
Thank you for your review, I totally agree that the CSVExampleGen should return "INT" type for the `pickup_sensus_tract`. Great work!

PR #1551, opened 2020-03-29
Fix two graphs with mismatched name and content

xyuzh:
Trainer and Evaluator graph in index page may need to swap


xyuzh:
@googlebot I signed it!

xyuzh:
@googlebot I signed it!

PR #1667, opened 2020-04-22
Refreshes the contributing.md.

zhitaoli:
Also included:
- Fix some tests which fails when executed externally, mostly due to
usage of testdata before CWD change, or required environment variables.
- Refreshes test dependency and remove unnecessary ones.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,33 +21,54 @@ All submissions, including submissions by project members, require review.
 We use GitHub pull requests for this purpose. Consult GitHub Help for more
 information on using pull requests.
 
+Proper code review should include proper unit tests with them. Please see
+[Testing Conventions](##Testing Conventions) for our conventions on how to write proper unit tests.
+
+TODO(zhitaoli): Configure a CI/CD on Github so that each PR submission is tested
+before merged.
+
 ## Community Guidelines
 This project follows[Google's Open Source Community Guidelines](
 https://opensource.google.com/conduct/).
 
 # Contributing Guidelines
+At this point, TFX only supports Python 3 on Linux and MacOS. Please use one of
+these operation system for development and testing.
 
-## Running Unit Tests
+If Python 3.5 is used, our usage of type hints requires at least 3.5.3.
 
-We recommend using virtualenv and Python 3.6 and above for development. You can
-invoke all Python unit tests:
+## Testing Conventions
+All python unit tests in this repo is based on Tensorflow's [tf.test.TestCase](
+https://www.tensorflow.org/api_docs/python/tf/test/TestCase), which is a subclass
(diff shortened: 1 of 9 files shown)


CONTRIBUTING.md
```diff
@@ -21,33 +21,54 @@ All submissions, including submissions by project members, require review.
...
 
 ## Testing local change
 To test local change, you will need to install code into virtualenv in editable
 mode:
 
-```
+```shell
 pushd <your_source_dir>
-pip install -e .
+pip install -e .[all]   # the [all] suffix includes additional packages for test
```
jiyongjung0, line 56: Because [all] installs python-snappy as well, it [requires snappy C library as well](https://github.com/andrix/python-snappy#dependencies). I thought that it would be more convenient if we don't have such a dependency.

So I'd like to propose a new "extra" configuration called "test". 
For example: https://github.com/tensorflow/tfx/blob/b0e93a3817bdba23e02d69c3695f3f108af82913/setup.py#L127

CONTRIBUTING.md
```diff
@@ -21,33 +21,54 @@ All submissions, including submissions by project members, require review.
...
+
+## Running Unit Tests
+
+At this point all unit tests are safe to run externaly. We are working on
+porting the end to end tests.
+
+Each test can just be invoked with `python`. To invoke all unit tests:
+
+```shell
+find . -name '*_test.py' | grep -v e2e | xargs -I {} python {}
```
jiyongjung0, line 67: I think that we can stick with `pytest` to run all unit tests. (Although I deleted an 'end_to_end' marker last time 😅 )
This is a command what I'm using.
`pytest --ignore-glob='*e2e*' --ignore-glob='*integration*' --ignore=tfx/examples tfx`

And I think that it would be better to include how to run unit tests for a few files. For example, files that contributors changed or added.

jiyongjung0: I found that pytest doesn't work well abseil and it is recommend to run the test using __main__() or `bazel`. ([related issue](https://github.com/abseil/abseil-py/issues/101))
Let me resolve this for now.

CONTRIBUTING.md
jiyongjung0: Per #1679, we can rely on `'all'`.

tfx/tools/cli/handler/beam_handler_test.py
```diff
@@ -357,7 +357,7 @@ def testCreateRun(self):
     handler = beam_handler.BeamHandler(flags_dict)
     with self.captureWritesToStream(sys.stdout) as captured:
       handler.create_run()
-    self.assertIn("[%s, '%s']" % (sys.executable, self.pipeline_path),
+    self.assertIn("['%s', '%s']" % (sys.executable, self.pipeline_path),
```
jiyongjung0, line 360: There is an issue with this if sys.executable is None or something non-string.
```suggestion
    self.assertIn("%s" % [sys.executable, self.pipeline_path],
```

PR #1685, opened 2020-04-25
Enable bigquery example gen to use multi process by using FnApiWorker.

zhitaoli:


zhitaoli:
Will generate a PR separately instead.

PR #1688, opened 2020-04-27
Create directory of value file for ValueArtifact if it does not exist.

wendy2003888:
Fix file or directory NotFoundError in standard_artifacts.String().write()
This error may also occurs when calling the value setter.

### DIFF
--- a/tfx/types/artifact.py
+++ b/tfx/types/artifact.py
@@ -488,8 +488,10 @@ class ValueArtifact(Artifact):
 
   def write(self, value):
     serialized_value = self.encode(value)
-    tf.io.gfile.GFile(os.path.join(self.uri, self.__class__.VALUE_FILE),
-                      'wb').write(serialized_value)
+    value_file_path = os.path.join(self.uri, self.__class__.VALUE_FILE)
+    if not tf.io.gfile.exists(value_file_path):
+      tf.io.gfile.makedirs(os.path.dirname(value_file_path))
+    tf.io.gfile.GFile(value_file_path, 'wb').write(serialized_value)
 
   @property
   def value(self):


numerology:
Hi @wendy2003888 thanks for the fix.

Before I approve this PR, do you mind elaborating on under which condition you observe the `NotFoundError`? The directory creation should be done as part of artifact preparation as in https://github.com/tensorflow/tfx/blob/40d2a28e7b892910902d64c8ad5b1a51e519e3b2/tfx/components/base/base_driver.py#L39 so when trying to write out to a non-existing dir, raising an error should be the intended behavior IMO.

wendy2003888:
Hi Jiaxiao,
I used standard_artifacts.String() as an input of a head custom component. 
If this is not a potential use case, feel free to close this PR.
I can create the directory before the pipeline starts.

numerology:
> Hi Jiaxiao,
> I used standard_artifacts.String() as an input of a head custom component.
> If this is not a potential use case, feel free to close this PR.
> I can create the directory before the pipeline starts.

This is actually a good point regarding usability. 

Currently to introduce an artifact to the TFX pipeline, we have two options:

1. [Importer](https://github.com/tensorflow/tfx/blob/40d2a28e7b892910902d64c8ad5b1a51e519e3b2/tfx/examples/chicago_taxi_pipeline/taxi_pipeline_importer.py#L88), which requires the artifact uri;
2. [ExampleGen](https://github.com/tensorflow/tfx/blob/40d2a28e7b892910902d64c8ad5b1a51e519e3b2/tfx/examples/chicago_taxi_pipeline/taxi_pipeline_importer.py#L82), which currently only works with `ExternalArtifact` type.

Both Importer and ExampleGen handle the artifact preparation/publish properly in their driver.
In your case the current 'blessed' way is to use an Importer I believe.

I will close this PR shortly but feel free to open an issue if you see fit. Thanks!

PR #1731, opened 2020-05-05
Add support for BigQuery arrays in BigQueryExampleGen and update test

domedagen:
When trying to construct a pipeline having an array as input (such as a time series) you'll get the error 
`TypeError: [20.592889320938852, 20.514573172830033, 19.76807155325737, 20.467801775051797, 20.5308510931856, 20 has type list, but expected one of: int, long, float [while running 'InputToSerializedExample/InputSourceToExample/ToTFExample']`

This is due to incomplete checks in BigQueryExampleGen as the schema type will still be float even if the value is a list (repeatable)

`SchemaField('history', 'FLOAT', 'REPEATED', None, ())`

This is a quick fix that adds support for arrays of the already supported types. You might want to reconsider the general parsing of BigQuery rows, but that seem to require a better understanding of both BigQuery and TFX.


domedagen:
@jyzhao was mentioned as a TODO here

tfx/components/example_gen/big_query_example_gen/executor_test.py
```diff
@@ -64,9 +67,13 @@ def setUp(self):
     # Mock BigQuery result schema.
     self._schema = [
         bigquery.SchemaField('i', 'INTEGER', mode='REQUIRED'),
+        bigquery.SchemaField('i2', 'INTEGER', mode='REQUIRED'),
```
1025KB, line 70: quick question, when BigQuery returns a list, is the scheme still "INTEGER"

1025KB:
> @jyzhao was mentioned as a TODO [here](https://github.com/tensorflow/tfx/blob/master/tfx/components/example_gen/big_query_example_gen/executor.py#L60).
> 
> > TODO(jyzhao): support more types.
> 
> I guess this is a type of type support?

Yep I think to support list, you need to add more types here

tfx/components/example_gen/big_query_example_gen/executor_test.py
domedagen: I think so. I don't exactly know where bigquery.SchemaField came from, but if you study the results.schema you find `SchemaField('array_agg', 'INTEGER', 'REPEATED', None, ())`. It could be that `mode='REPEATED'` though.

```
query = """
SELECT ARRAY_AGG(x) AS array_agg
FROM UNNEST([2, 1, -2, 3, -2, 1, 2]) AS x;
"""


client = bigquery.Client()
# Dummy query to get the type information for each field.
query_job = client.query(query)
results = query_job.result()
type_map = {}

for field in results.schema:
    print(field)
```

domedagen: It can apparently also be NULLABLE `SchemaField('array_agg', 'STRING', 'NULLABLE', None, ())`
```
query = """
SELECT FORMAT("%T", ARRAY_AGG(x)) AS array_agg
FROM UNNEST([2, 1, -2, 3, -2, 1, 2]) AS x;
"""


client = bigquery.Client()
# Dummy query to get the type information for each field.
query_job = client.query(query)
results = query_job.result()
type_map = {}

for field in results.schema:
    print(field)
```

1025KB: > It can apparently also be NULLABLE `SchemaField('array_agg', 'STRING', 'NULLABLE', None, ())`

handling None is tricky, as different users might want different behaviors (empty feature or feature not exist), if data contains None, using ImportExampleGen can provide the desired behavior as the tf.Examples is created by users

1025KB: > I think so. I don't exactly know where bigquery.SchemaField came from, but if you study the results.schema you find `SchemaField('array_agg', 'INTEGER', 'REPEATED', None, ())`. [In the current implementation](https://github.com/tensorflow/tfx/blob/b66ad84e9fd216b78bc72ea5bc731d027a3b3f4d/tfx/components/example_gen/big_query_example_gen/executor.py#L42) this is how the type is derived.
> 
> It could be that `mode='REPEATED'` though. It is not intuitively obvious that a repeated/array field can't be nullable or required. [This one](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#TableFieldSchema.FIELDS.mode) however says it one of them.

Is that possible to check this REPEATED instead of isinstance(xx, list)?

domedagen: You mean NULLABLE and not None? The None is likely the description and the empty tuple is likely fields ([docs](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#TableFieldSchema.FIELDS.mode)).

In this case NULLABLE seems to refer to that the entire array could be missing (not that the array is empty.
```
SELECT ARRAY_AGG(x) AS array_agg
FROM UNNEST([]) AS x;
```
`SchemaField('array_agg', 'INTEGER', 'REPEATED', None, ())`

The schema is lost at that point in the current structure of the program (`__init__` vs `RowToExample`). We could however assign it the type to "REPEATED_INTEGER" or something. The question is that is an improvement.

Another question is also, does BigQuery allow for array of arrays and other corner cases?

1025KB: Scheme is returned by BigQuery, we can not set it (except in test, we fake the BigQuery result)

domedagen: @1025KB : Sorry, I was vague and I rephrased it now. I meant that we could create a new type in our `_type_map`

I tried the solution of handling all the arrays in one place, but I don't think it is better than what we have now.

```
    for key, values in instance.items():
      data_type = self._type_map[key]
      if values is None:
        feature[key] = tf.train.Feature()
      else:
        # Handle arrays and single values in the same way
        if not isinstance(values, list):
          values = [values]
        elif data_type in ('INTEGER', 'BOOLEAN'):
          feature[key] = tf.train.Feature(
              int64_list=tf.train.Int64List(value=values))
        elif data_type == 'FLOAT':
          feature[key] = tf.train.Feature(
              float_list=tf.train.FloatList(value=values))
        elif data_type == 'STRING':
          feature[key] = tf.train.Feature(
              bytes_list=tf.train.BytesList(value=[tf.compat.as_bytes(value) for value in values]))
        else:
          # TODO(jyzhao): support more types.
          raise RuntimeError(
              'BigQuery column type {} is not supported.'.format(data_type))
    return tf.train.Example(features=tf.train.Features(feature=feature))
```

tfx/components/example_gen/big_query_example_gen/executor_test.py
```diff
@@ -64,9 +67,13 @@ def setUp(self):
     # Mock BigQuery result schema.
     self._schema = [
         bigquery.SchemaField('i', 'INTEGER', mode='REQUIRED'),
+        bigquery.SchemaField('i2', 'INTEGER', mode='REPEATED'),
         bigquery.SchemaField('b', 'BOOLEAN', mode='REQUIRED'),
         bigquery.SchemaField('f', 'FLOAT', mode='REQUIRED'),
+        bigquery.SchemaField('f2', 'FLOAT', mode='REPEATED'),
         bigquery.SchemaField('s', 'STRING', mode='REQUIRED'),
+        bigquery.SchemaField('s2', 'STRING', mode='NULLABLE'),
```
1025KB, line 75: REPEATED?

tfx/components/example_gen/big_query_example_gen/executor.py
```diff
@@ -51,13 +51,16 @@ def RowToExample(self, instance: Dict[Text, Any]) -> tf.train.Example:
         feature[key] = tf.train.Feature()
       elif data_type in ('INTEGER', 'BOOLEAN'):
         feature[key] = tf.train.Feature(
-            int64_list=tf.train.Int64List(value=[value]))
+            int64_list=tf.train.Int64List(value=value if isinstance(value, list) else [value]))
```
1025KB, line 54: can we check the REPEATED instead of isinstance(value, list)?

tfx/components/example_gen/big_query_example_gen/executor_test.py
```diff
@@ -64,9 +67,13 @@ def setUp(self):
     # Mock BigQuery result schema.
     self._schema = [
         bigquery.SchemaField('i', 'INTEGER', mode='REQUIRED'),
+        bigquery.SchemaField('i2', 'INTEGER', mode='REPEATED'),
```
1025KB, line 70: quick question, when you run a real BigQuery that returns an array, its scheme will looks like this?

tfx/components/example_gen/big_query_example_gen/executor_test.py
domedagen: I'm afraid it can actually be NULLABLE as well. [Example](https://github.com/tensorflow/tfx/pull/1731#discussion_r421064061)

domedagen: In the current implementation is however ignored so it doesn't matter, but future solutions should be aware of this.

tfx/components/example_gen/big_query_example_gen/executor.py
domedagen: Not without redesigning it.

We get the schema in the __init__ and then we only have the row here, which doesn't say anything about the [original BQ type](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.table.Row.html).

tfx/components/example_gen/big_query_example_gen/executor_test.py
domedagen: Yes, you can try it yourself [here](https://github.com/tensorflow/tfx/pull/1731#discussion_r421053710).

tfx/components/example_gen/big_query_example_gen/executor_test.py
domedagen: I also updated the test so that the mode is more realistic

Could you elaborate on ImportExampleGen and what you think would be the suitable thing to do if we get an empty array. I guess that if the array is NULL, it would be handled the same way as when an integer is null.

tfx/components/example_gen/big_query_example_gen/executor_test.py
1025KB: for this test, since 's2': ['abc', 'def'], it should be REPEATED? or you mean it might be None here?

domedagen: Exactly! It is not NULL, but it is NULLABLE as in could be NULL. See example [here](https://github.com/tensorflow/tfx/pull/1731#discussion_r421064061).

1025KB: NULLABLE means 's2' could be missing or 's2' could be an empty list

1025KB: for 's2': ['abc', 'def'], it is REPEATED
for 's2': "['abc', 'def']", it might be REQUIRED or NULLABLE

domedagen: Sorry for being so slow. You're correct, my example don't actually say anything about NULLABLE or REPEATED. If I had studied [FORMAT](https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions#format_string) more closely, rather than just copy-pasted it, it actually just converts any format into a string. 

If you replace the example query with `SELECT CAST(NULL AS ARRAY<String>) AS x UNION ALL
SELECT ['str1', 'str2'] AS x`
SchemaField('x', 'STRING', 'REPEATED', None, ())

This proves that it is REPEATED eventhough it is NULL. I'll update it!

1025KB:
If https://github.com/tensorflow/tfx/pull/1760 get merged, this one will automatically get merged too. Thanks!

PR #1751, opened 2020-05-07
Fix kwargs for AirflowPipelineConfig

casassg:
Whenever you need to change  an AirflowPipelineConfig (other than `airflow_dag_config`) then its passed as a dict into `support_component_launchers`. 

This unrolls the kwargs argument such that it overwrites the correct arguments.


casassg:
@googlebot I signed it!

Issue #1497, opened 2020-03-16
ImportError: cannot import name 'CsvExampleGen' on r0.21. / python 3.6.9 / tf 1.14


jason-brian-anderson:
I'm hoping to deploy TFX in product at Garmin and am running into some issues with the [Airflow tutorial](https://www.tensorflow.org/tfx/tutorials/tfx/airflow_workshop) 

i'm wondering if my TF version is too old?  i just ran 

`/home/tfx/tfx/tfx/examples/airflow_workshop/setup/setup_demo.sh 
`

which runs:

`printf "${GREEN}Installing TensorFlow${NORMAL}\n"

pip install tensorflow==1.14.0`

And the `README.md` testing framework compatibility tables states that r 0.21.0 should use TF 1.15.0 / 2.1

Anyway, here's my env:

```
(tfx-env) tfx@lambda-quad:~/tfx$ which python
/home/tfx/tfx-env/bin/python

(tfx-env) tfx@lambda-quad:~/tfx$ python -V
Python 3.6.9
(tfx-env) tfx@lambda-quad:~/tfx$ 

(tfx-env) tfx@lambda-quad:~/tfx$ git status
HEAD detached at origin/r0.21
nothing to commit, working tree clean


(tfx-env) tfx@lambda-quad:~/tfx$ python -c 'import tensorflow as tf; print(tf.__version__)'
<excluded a bunch of warnings...>
1.14.0

(tfx-env) tfx@lambda-quad:~/tfx$ git checkout -f origin/r0.21
Previous HEAD position was 711a9d8 Pin versions of airflow, flask and Werkzeug in setup_demo.sh
HEAD is now at 8f5a13f Merge pull request #1279 from dhruvesh09/master
(tfx-env) tfx@lambda-quad:~/tfx$ airflow webserver -p 8080
[2020-03-16 13:28:56,928] {__init__.py:51} INFO - Using executor SequentialExecutor
  ____________       _____________
 ____    |__( )_________  __/__  /________      __
____  /| |_  /__  ___/_  /_ __  /_  __ \_ | /| / /
___  ___ |  / _  /   _  __/ _  / / /_/ /_ |/ |/ /
 _/_/  |_/_/  /_/    /_/    /_/  \____/____/|__/
[2020-03-16 13:28:57,185] {__init__.py:305} INFO - Filling up the DagBag from /home/tfx/airflow/dags
[...]

PR #1849, opened 2020-05-26
[Airflow tutorial] Remove Step 1's checkout out of r0.21

atn832:
Because it causes Step 1's `setup_demo.sh` to fail. See #1497. It happened to me today as well.

```shell
$ python -V
Python 3.6.8
```


chongkong:
Hi Ahn Tuan! I think you shall send CL to the Piper rather than from Github. Do you have specific reason?

PR #1850, opened 2020-05-26
[Airflow tutorial] Update Airflow dependency to 1.10.10

atn832:
Because Airflow 1.10.9 won't run in step 2, when running `airflow webserver -p 8080`.

Airflow 1.10.9 has a problem with its SQLAlchemy dependency. They fixed it in 1.10.10. See https://github.com/apache/airflow/issues/8211.


chongkong:
Hi Ahn Tuan! I think you shall send CL to the Piper rather than from Github. Do you have specific reason?

PR #2004, opened 2020-06-16
Added support for beam_pipeline_args to InteractiveContext constructor

kshivvy:
Prevents having to specify beam_pipeline_args every time run() is called, will default to value utilized in constructor if none is passed in.


kshivvy:
@charlesccychen 
@yifanmai

codesue:
Created issue https://github.com/tensorflow/tfx/issues/2371 to track this PR. Commenting to remove the stale label.

PR #2029, opened 2020-06-22
cifar10 transfer learning example

Deanplayerljx:


davidzats-eng:
@1025KB I understand the concern about too many example types. However, this example demonstrates how TFX can perform high-quality image classification on real-world datasets. As such, I don't think the MNSIT dataset is appropriate.

zhitaoli:
@davidzats-eng Can you make sure you set yourself as owner of this example once pulled in? Can we also think about a secondary endorser (who is familiar with the modeling technique or so)?

davidzats-eng:
@zhitaoli Ack will do.

tfx/examples/cifar10/cifar10_pipeline_native_keras.py
```diff
@@ -0,0 +1,180 @@
...
+
+  def _create_trainer(module_file, instance_name):
+    return Trainer(
+        module_file=module_file,
+        custom_executor_spec=executor_spec.ExecutorClassSpec(GenericExecutor),
+        examples=transform.outputs['transformed_examples'],
+        transform_graph=transform.outputs['transform_graph'],
+        schema=schema_gen.outputs['schema'],
+        train_args=trainer_pb2.TrainArgs(num_steps=10000),
+        eval_args=trainer_pb2.EvalArgs(num_steps=156),
```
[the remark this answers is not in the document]
Deanplayerljx: The validation set has 10000 examples and batch size is 64. I want to evaluate on the whole validation set, so I set the number of steps 10000/64 ~= 156.

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,192 @@
...
+      tf.keras.layers.InputLayer(
+          input_shape=(224, 224, 3), name=transformed_name(IMAGE_KEY)),
+      base_model,
+      tf.keras.layers.Dropout(0.3),
+      tf.keras.layers.Dense(10, activation='softmax')
+  ])
+
+  model.compile(
+      loss='sparse_categorical_crossentropy',
+      optimizer=tf.keras.optimizers.RMSprop(lr=0.0015),
```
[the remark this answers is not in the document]
Deanplayerljx: I didn't tune this because I found it worked well. This is the same learning rate as the one in mnist example.

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,192 @@
...
+
+  Args:
+    inputs: map from feature keys to raw not-yet-transformed features.
+
+  Returns:
+    Map from string feature key to transformed feature operations.
+  """
+  outputs = {}
+
+  image_features = tf.map_fn(lambda x: tf.io.decode_png(x[0], channels=3), inputs[IMAGE_KEY], dtype=tf.uint8)
```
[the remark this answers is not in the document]
Deanplayerljx: The inputs[IMAGE_KEY] is a string tensor of rank 1, i.e. [string], which was loaded from the TF record file prepared by the TF Datasets. the tf.io.decode_png function expects a string tensor of rank 0, so I pulled the string value out.

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,192 @@
...
+  image_features = tf.cast(image_features, tf.float32)
+
+  # The MobileNet we use was trained on ImageNet, which has image size 224 x 224.
+  # We resize CIFAR10 images to match that size
+  image_features = tf.image.resize(image_features, [224, 224])
+
+  image_features = tf.ensure_shape(image_features, (None, 224, 224, 3))
```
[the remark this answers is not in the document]
Deanplayerljx: Here it's just a check, because I did not use tft functions. It will be necessary if we want to use tf.map_fn and tft functions together, since tft functions require "all the dimensions except batch dimension should be known", and the tf.map_fn will make some other dimension to be None.

tfx/examples/cifar10/cifar10_pipeline_native_keras.py
```diff
@@ -0,0 +1,183 @@
...
+  eval_config = tfma.EvalConfig(
+      model_specs=[tfma.ModelSpec(label_key='label')],
+      slicing_specs=[tfma.SlicingSpec()],
+      metrics_specs=[
+          tfma.MetricsSpec(metrics=[
+              tfma.MetricConfig(
+                  class_name='SparseCategoricalAccuracy',
+                  threshold=tfma.config.MetricThreshold(
+                      value_threshold=tfma.GenericValueThreshold(
+                          lower_bound={'value': 0.8}),
```
davidzats-eng, line 127: What value for accuracy are we getting here? Can / should this be adjusted?

tfx/examples/cifar10/cifar10_pipeline_native_keras.py
```diff
@@ -0,0 +1,183 @@
...
+                      change_threshold=tfma.GenericChangeThreshold(
+                          direction=tfma.MetricDirection.HIGHER_IS_BETTER,
+                          absolute={'value': -1e-10})))
```
davidzats-eng, line 130: This seems extremely tight, how about -1e-3 or so? By the way, where did this number come from? Because it seems too tight for most use-cases.

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,204 @@
...
+                            _TRAIN_BATCH_SIZE)
+  eval_dataset = _input_fn(fn_args.eval_files, tf_transform_output,
+                           _EVAL_BATCH_SIZE)
+
+  mirrored_strategy = tf.distribute.MirroredStrategy()
+  with mirrored_strategy.scope():
+    model = _build_keras_model()
+
+  steps_per_epoch = _TRAIN_DATA_SIZE / _TRAIN_BATCH_SIZE
+  epochs = int(fn_args.train_steps / steps_per_epoch)
```
davidzats-eng, line 188: @1025KB is this the right thing to do here? IIRC there were issues with multi-epoch training.

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,204 @@
...
+
+from typing import List, Text
+
+import absl
+import tensorflow as tf
+import tensorflow_transform as tft
+
+from tfx.components.trainer.executor import TrainerFnArgs
+
+# cifar10 dataset has 50000 train records, and 10000 val records
```
davidzats-eng, line 33: s/val/eval?

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,204 @@
...
+
+def _freeze_model_by_percentage(model: tf.keras.Model,
+                                percentage: float):
+  """Freeze part of the model based on specified percentage
+
+  Args:
+    model: The keras model need to be partially frozen
+    percentage: the percentage of layers to freeze
+  """
+  percentage = max(0.0, percentage)
```
davidzats-eng, line 103: It may be worthwhile to actually raise an error instead of hiding it with the max / min. Otherwise the user may get the wrong impression.

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,204 @@
...
+    layer.trainable = False
+
+def _build_keras_model() -> tf.keras.Model:
+  """Creates a MobileNet model pretrained on ImageNet for finetuning
+
+  Returns:
+    A Keras Model.
+  """
+  tf.keras.backend.set_image_data_format('channels_last')
+  base_model = tf.keras.applications.MobileNet(
```
davidzats-eng, line 119: question: why MobileNet vs MobileNetv2?

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,204 @@
...
+      break
+    layer.trainable = False
+
+def _build_keras_model() -> tf.keras.Model:
+  """Creates a MobileNet model pretrained on ImageNet for finetuning
+
+  Returns:
+    A Keras Model.
+  """
+  tf.keras.backend.set_image_data_format('channels_last')
```
davidzats-eng, line 118: The default seems to be 'channels_last' https://github.com/tensorflow/tensorflow/blob/2b96f3662bd776e277f86997659e61046b56c315/tensorflow/python/keras/backend_config.py#L29

So, do we need this line?

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,204 @@
...
+      input_shape=(224, 224, 3), include_top=False, weights='imagenet',
+      pooling='avg')
+
+  _freeze_model_by_percentage(base_model, 0.9)
+
+  model = tf.keras.Sequential([
+      tf.keras.layers.InputLayer(
+          input_shape=(224, 224, 3), name=transformed_name(IMAGE_KEY)),
+      base_model,
+      tf.keras.layers.Dropout(0.3),
```
davidzats-eng, line 129: I am assuming that if you are putting in a large dropout that you are seeing overfitting. If so, then perhaps data augmentation using tools such as  tf.image.random_crop, tf.image.random_brightnes, .. is the way to go.

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,204 @@
...
+
+  Returns:
+    A Keras Model.
+  """
+  tf.keras.backend.set_image_data_format('channels_last')
+  base_model = tf.keras.applications.MobileNet(
+      input_shape=(224, 224, 3), include_top=False, weights='imagenet',
+      pooling='avg')
+
+  _freeze_model_by_percentage(base_model, 0.9)
```
davidzats-eng, line 123: My understanding is that the best practice is to train for a bit using the new final layer before unfreezing any part of the model to ensure that the new weights do not pollute the existing model. Can we please do that here?

tfx/examples/cifar10/cifar10_utils_native_keras.py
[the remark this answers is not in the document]
davidzats-eng: Can you write a comment about that?

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,204 @@
...
+  image_features = tf.cast(image_features, tf.float32)
+
+  # The MobileNet we use was trained on ImageNet, which has image size 224 x 224.
+  # We resize CIFAR10 images to match that size
+  image_features = tf.image.resize(image_features, [224, 224])
+
+  image_features = tf.map_fn(tf.keras.applications.mobilenet.preprocess_input,
+                             image_features, dtype=tf.float32)
+
+  outputs[transformed_name(IMAGE_KEY)] = (image_features)
```
davidzats-eng, line 162: Why the parents around (image_features)?

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,204 @@
...
+  outputs = {}
+
+  image_features = tf.map_fn(lambda x: tf.io.decode_png(x[0], channels=3), inputs[IMAGE_KEY], dtype=tf.uint8)
+  image_features = tf.cast(image_features, tf.float32)
+
+  # The MobileNet we use was trained on ImageNet, which has image size 224 x 224.
+  # We resize CIFAR10 images to match that size
+  image_features = tf.image.resize(image_features, [224, 224])
+
+  image_features = tf.map_fn(tf.keras.applications.mobilenet.preprocess_input,
```
davidzats-eng, line 159: Nit: Can we combine the two map functions into one? It might be more readable?

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,204 @@
...
+    Map from string feature key to transformed feature operations.
+  """
+  outputs = {}
+
+  image_features = tf.map_fn(lambda x: tf.io.decode_png(x[0], channels=3), inputs[IMAGE_KEY], dtype=tf.uint8)
+  image_features = tf.cast(image_features, tf.float32)
+
+  # The MobileNet we use was trained on ImageNet, which has image size 224 x 224.
+  # We resize CIFAR10 images to match that size
+  image_features = tf.image.resize(image_features, [224, 224])
```
davidzats-eng, line 157: There are multiple ways of resizing. Why do we believe this provides the best quality for our dataset?

davidzats-eng: See https://www.tensorflow.org/api_docs/python/tf/image/ResizeMethod for options.

tfx/examples/cifar10/README.md
```diff
@@ -0,0 +1,54 @@
+
+# CIFAR-10 Transfer Learning Example
```
davidzats-eng, line 2: lets also add a unit test like so:
https://github.com/tensorflow/tfx/blob/master/tfx/examples/mnist/mnist_pipeline_native_keras_e2e_test.py

tfx/examples/cifar10/cifar10_utils_native_keras.py
Deanplayerljx: hmmm i think the usual terms are train/validation/test sets?

tfx/examples/cifar10/cifar10_pipeline_native_keras.py
Deanplayerljx: Sure we can do that, i copied it from the iris example i remember

tfx/examples/cifar10/cifar10_pipeline_native_keras.py
Deanplayerljx: it was about 0.9, lower_bound of 0.8 should be fine

tfx/examples/cifar10/cifar10_pipeline_native_keras.py
```diff
@@ -0,0 +1,183 @@
...
+from tfx.proto import example_gen_pb2
+from tfx.utils.dsl_utils import external_input
+
+_pipeline_name = 'cifar10_native_keras'
+
+# This example assumes that CIFAR10 train set data is stored in
+# ~/cifar10/data/train, test set data is stored in ~/cifar10/data/test, and
+# the utility function is in ~/cifar10. Feel free to customize as needed.
+_cifar10_root = os.path.join(os.environ['HOME'], 'cifar10')
+_data_root = os.path.join(_cifar10_root, 'data')
```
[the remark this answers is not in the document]
Deanplayerljx: We can do this but I'm not sure if the model will have decent accuracy when just training on 100 images..

tfx/examples/cifar10/cifar10_utils_native_keras.py
Deanplayerljx: MobileNet has a simpler architecture and i thought it be more suitable as a tutorial example.

tfx/examples/cifar10/cifar10_utils_native_keras.py
Deanplayerljx: I remember I got an error about this, will double check

tfx/examples/cifar10/cifar10_utils_native_keras.py
Deanplayerljx: hmm I think the default option (bilinear pooling) is the most common practice. It worked well as the model got ~90% acc on validation set.

tfx/examples/cifar10/cifar10_utils_native_keras.py
Deanplayerljx: In my understanding data augmentation and dropout (just like batchnorm between convolutional layers) are two parallel techniques for improving model's generalization ability, one at data level and another at model level. They are both fine as long as our model got decent accuracy?

tfx/examples/cifar10/cifar10_utils_native_keras.py
Deanplayerljx: I have tried that setting, i.e. train new layers for 6 epochs and unfreeze all layers for another 6 epochs. But it didn't show improvement... Should I still add it?

tfx/examples/cifar10/cifar10_pipeline_native_keras.py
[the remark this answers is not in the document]
davidzats-eng: its more to have a working copy. you can document what the best parameters are for the full dataset, etc.

tfx/examples/cifar10/cifar10_pipeline_native_keras.py
davidzats-eng: @1025KB Should we change our examples? This seems to tight for most use-cases.

tfx/examples/cifar10/cifar10_utils_native_keras.py
davidzats-eng: val was unclear to me. so then s/val/validation :)

tfx/examples/cifar10/cifar10_utils_native_keras.py
davidzats-eng: its good to highlight best practices so yes. and maybe write a comment that this part is "optional".

tfx/examples/cifar10/cifar10_utils_native_keras.py
davidzats-eng: yes, you are right, they are orthogonal. I was just thinking there was opportunity for improving accuracy with data augmentation.

tfx/examples/cifar10/cifar10_utils_native_keras.py
davidzats-eng: lets push to see how high we can go. Just because we get ~90% accuracy on an easy dataset doesn't mean that we can't push for more.

tfx/examples/cifar10/cifar10_pipeline_native_keras.py
```diff
@@ -0,0 +1,183 @@
...
+# Sqlite ML-metadata db path.
+_metadata_path = os.path.join(_tfx_root, 'metadata', _pipeline_name,
+                              'metadata.db')
+
+def _create_pipeline(pipeline_name: Text, pipeline_root: Text, data_root: Text,
+                     module_file: Text,
+                     serving_model_dir: Text,
+                     metadata_path: Text,
+                     direct_num_workers: int) -> pipeline.Pipeline:
+  """Implements the cifar10 image classification pipeline using TFX."""
```
liamcrawford, line 74: nit: standardize capitalization in documentation (i.e., "CIFAR10" vs. "cifar10")

liamcrawford: ditto in other files

tfx/examples/cifar10/meta_data_writer.py
```diff
@@ -0,0 +1,208 @@
...
+
+FLAGS = flags.FLAGS
+
+
+def define_flags():
+  flags.DEFINE_string("model_file", None,
+                      "Path and file name to the TFLite model file.")
+  flags.DEFINE_string("label_file", None, "Path to the label file.")
+  flags.DEFINE_string("export_directory", None,
+                      "Path to save the TFLite model files with metadata.")
```
davidzats-eng, line 41: Do we handle the case where the input and output directories are the same? If not, perhaps an explicit error message would be good.

tfx/examples/cifar10/meta_data_writer.py
```diff
@@ -0,0 +1,208 @@
...
+
+  # Copies model_file to export_path.
+  tf.io.gfile.copy(model_file, export_model_path, overwrite=True)
+
+  # Generate the metadata objects and put them in the model file
+  populator = MetadataPopulatorForImageClassifier(
+      export_model_path, _MODEL_INFO.get(model_basename), FLAGS.label_file)
+  populator.populate()
+
+  # Validate the output model file by reading the metadata and produce
```
davidzats-eng, line 189: Is the json file useful? If so, why is it "validation"? If not, then perhaps have a validation flag and perhaps store the json file in tmpdir instead of placing it with the export.

tfx/examples/cifar10/meta_data_writer.py
```diff
@@ -0,0 +1,208 @@
...
+  # a json file with the metadata under the export path
+  displayer = _metadata.MetadataDisplayer.with_model_file(export_model_path)
+  export_json_file = os.path.join(FLAGS.export_directory,
+                                  os.path.splitext(model_basename)[0] + ".json")
+  json_file = displayer.get_metadata_json()
+  with open(export_json_file, "w") as f:
+    f.write(json_file)
+
+  print("Finished populating metadata and associated file to the model:")
```
davidzats-eng, line 198: use logging instead of prints.

tfx/examples/cifar10/meta_data_writer.py
```diff
@@ -0,0 +1,208 @@
...
+    """Creates the metadata for an image classifier."""
+
+    # Creates model info.
+    model_meta = _metadata_fb.ModelMetadataT()
+    model_meta.name = self.model_info.name
+    model_meta.description = ("Identify the most prominent object in the "
+                              "image from a set of %d categories." %
+                              self.model_info.num_classes)
+    model_meta.version = self.model_info.version
+    model_meta.author = "TensorFlow"
```
davidzats-eng, line 102: what happens when someone not from the tensorflow generates the metadata?

tfx/examples/cifar10/meta_data_writer.py
```diff
@@ -0,0 +1,208 @@
...
+    input_meta.processUnits = [input_normalization]
+    input_stats = _metadata_fb.StatsT()
+    input_stats.max = [self.model_info.image_max]
+    input_stats.min = [self.model_info.image_min]
+    input_meta.stats = input_stats
+
+    # Creates output info.
+    output_meta = _metadata_fb.TensorMetadataT()
+    output_meta.name = "probability"
+    output_meta.description = "Probabilities of the %d labels respectively." % \
```
davidzats-eng, line 136: instead of "\", wrap the multiline statement in parenthesis.

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,259 @@
...
+
+  model = tf.keras.Sequential([
+      tf.keras.layers.InputLayer(
+          input_shape=(224, 224, 3), name=transformed_name(IMAGE_KEY)),
+      base_model,
+      tf.keras.layers.Dropout(0.3),
+      tf.keras.layers.Dense(10, activation='softmax')
+  ])
+
+  model.compile(
```
davidzats-eng, line 130: It looks like this compilation is never used. Perhaps remove? Or perhaps remove the other one?

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,259 @@
...
+  """
+  base_model = tf.keras.applications.MobileNet(
+      input_shape=(224, 224, 3), include_top=False, weights='imagenet',
+      pooling='avg')
+
+  model = tf.keras.Sequential([
+      tf.keras.layers.InputLayer(
+          input_shape=(224, 224, 3), name=transformed_name(IMAGE_KEY)),
+      base_model,
+      tf.keras.layers.Dropout(0.3),
```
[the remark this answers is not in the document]
davidzats-eng: For example, the default dropout seems to be 0.001 when the top is included in the original mobilenet model: https://github.com/tensorflow/tensorflow/blob/0d7e40f0dba759a54a8ee4e4c2a96e05937e14c2/tensorflow/python/keras/applications/mobilenet.py#L125

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,259 @@
...
+  dataset = tf.data.experimental.make_batched_features_dataset(
+      file_pattern=file_pattern,
+      batch_size=batch_size,
+      features=transformed_feature_spec,
+      reader=_gzip_reader_fn,
+      label_key=transformed_name(LABEL_KEY))
+
+  return dataset
+
+def _freeze_model_by_percentage(model: tf.keras.Model,
```
davidzats-eng, line 91: There is also a model.trainable parameter lets make sure that it is not overriding the settings here. Can you please verify?

tfx/examples/cifar10/README.md
```diff
@@ -0,0 +1,83 @@
...
+mkdir test_whole
+mv ciar10/3.0.2/cifar10-train.tfrecord-00000-of-00001 train_whole/
+mv ciar10/3.0.2/cifar10-test.tfrecord-00000-of-00001 test_whole/
+```
+### Train the model
+Execute the pipeline python file :
+```
+python ~/cifar10/cifar_pipeline_native_keras.py
+```
+The trained model can be found at `~/cifar10/serving_model_lite`
```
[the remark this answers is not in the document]
Deanplayerljx: I thought renaming it would make it more informative. By default the name is "tflite". I can remove the renaming part if you thing it's unnecessary.

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,259 @@
...
+          input_shape=(224, 224, 3), name=transformed_name(IMAGE_KEY)),
+      base_model,
+      tf.keras.layers.Dropout(0.3),
+      tf.keras.layers.Dense(10, activation='softmax')
+  ])
+
+  model.compile(
+      loss='sparse_categorical_crossentropy',
+      optimizer=tf.keras.optimizers.RMSprop(lr=0.0015),
+      metrics=['sparse_categorical_accuracy'])
```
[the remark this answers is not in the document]
Deanplayerljx: sure I will add the data augmentation

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,259 @@
...
+  base_model = tf.keras.applications.MobileNet(
+      input_shape=(224, 224, 3), include_top=False, weights='imagenet',
+      pooling='avg')
+
+  model = tf.keras.Sequential([
+      tf.keras.layers.InputLayer(
+          input_shape=(224, 224, 3), name=transformed_name(IMAGE_KEY)),
+      base_model,
+      tf.keras.layers.Dropout(0.3),
+      tf.keras.layers.Dense(10, activation='softmax')
```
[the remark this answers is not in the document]
Deanplayerljx: I have tried three layers before and i remember it did not give improvement. The origin MobileNet also only has one Dense layer, and it seems it's a common practice to just add one Dense layer as classification layer, e.g. [The Tensorflow transfer learning example (MobileNet backbone)](https://www.tensorflow.org/tutorials/images/transfer_learning)

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,259 @@
...
+  with mirrored_strategy.scope():
+    model = _build_keras_model()
+
+  steps_per_epoch = _TRAIN_DATA_SIZE / _TRAIN_BATCH_SIZE
+  total_epochs = int(fn_args.train_steps / steps_per_epoch)
+  classifier_epochs = int(total_epochs / 2)
+  finetune_epochs = total_epochs - classifier_epochs
+
+  # Freeze the whole MobileNet backbone and train the top classifer only
+  base_model = model.get_layer('mobilenet_1.00_224')
```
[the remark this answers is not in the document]
Deanplayerljx: sure i can do that

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,259 @@
...
+  model.fit(
+      train_dataset,
+      epochs=classifier_epochs,
+      steps_per_epoch=steps_per_epoch,
+      validation_data=eval_dataset,
+      validation_steps=fn_args.eval_steps
+  )
+
+  # Optional
+  # Unfreeze the top MobileNet layers and finetune the whole model
```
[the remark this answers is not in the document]
Deanplayerljx: yes we are just fine-tuning the last 10%.

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,259 @@
...
+  base_model = model.get_layer('mobilenet_1.00_224')
+  _freeze_model_by_percentage(base_model, 0.9)
+  # We need to recompile the model because layer properties have changed
+  model.compile(
+      loss='sparse_categorical_crossentropy',
+      optimizer=tf.keras.optimizers.RMSprop(lr=0.0005),
+      metrics=['sparse_categorical_accuracy'])
+  model.summary(print_fn=absl.logging.info)
```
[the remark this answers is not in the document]
Deanplayerljx: I feel that it will be helpful for the users to know the model status after each recompile, as they will be able to see how many parameters are frozen. Maybe we can add a logging before each summary, e.g. "Start training the top classifier" and "Start fine-tuning the model" ?

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,259 @@
...
+  Args:
+    model: The keras model need to be partially frozen
+  Returns:
+    A Keras Model.
+  """
+  base_model = tf.keras.applications.MobileNet(
+      input_shape=(224, 224, 3), include_top=False, weights='imagenet',
+      pooling='avg')
+
+  model = tf.keras.Sequential([
```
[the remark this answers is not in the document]
Deanplayerljx: Sure I can check that

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,259 @@
...
+                      shape=[None, 224, 224, 3],
+                      dtype=tf.float32,
+                      name=transformed_name(IMAGE_KEY)
+                      ))
+  }
+
+  temp_saving_model_dir = os.path.join(fn_args.serving_model_dir, 'temp')
+  model.save(temp_saving_model_dir, save_format='tf', signatures=signatures)
+
+  tfrw = rewriter_factory.create_rewriter(
```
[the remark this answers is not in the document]
Deanplayerljx: hmmm the reason we do not add transform into the serving model is that in our pipeline transform takes encoded image strings as input, whereas for serving in MLKit the model's input will be image pixel values. I don't think there is a way to add transform into the serving model for our use case.

tfx/examples/cifar10/cifar10_utils_native_keras.py
```diff
@@ -0,0 +1,259 @@
...
+LABEL_KEY = 'label'
+
+def transformed_name(key):
+  return key + '_xf'
+
+def _gzip_reader_fn(filenames):
+  """Small utility returning a record reader that can read gzip'ed files."""
+  return tf.data.TFRecordDataset(filenames, compression_type='GZIP')
+
+def _get_serve_tf_examples_fn(model):
```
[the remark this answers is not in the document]
Deanplayerljx: I mentioned the reason of excluding transform in serving model in the above conversation. I think TFLite rewriter will not have any problem writing the model with transform. The thing here is that usually in serving, models will take image tensors directly, which is also indicated in the MNIST example. MLKit and MediaPipe also requires models to take image tensors directly.

tfx/examples/cifar10/meta_data_writer.py
```diff
@@ -0,0 +1,208 @@
...
+    populator.load_associated_files([self.label_file_path])
+    populator.populate()
+
+
+def main(_):
+  model_file = FLAGS.model_file
+  model_basename = os.path.basename(model_file)
+  if model_basename not in _MODEL_INFO:
+    raise ValueError(
+        "The model info for, {0}, is not defined yet.".format(model_basename))
```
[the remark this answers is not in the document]
Deanplayerljx: hmmm here it's not deriving the model type, it's trying to derive the whole name of the model, i.e. "cifar10.tflite" in our case.

tfx/examples/cifar10/meta_data_writer.py
Deanplayerljx: sounds good

tfx/examples/cifar10/meta_data_writer.py
Deanplayerljx: This is just the model's description, i.e. you can define it whatever you want and it will not have any side effects.

davidzats-eng: Sure, but maybe its best to have this be configurable vs hard-coded :)

tfx/examples/cifar10/meta_data_writer.py
Deanplayerljx: The json file serves as a human-readable file to tell user what their model's metadata looks like. I think it has some value for the user. It's "validation" in a sense that it verify the exported model's metadata can be read in properly. I can change the comment if you think it's not very appropriate.

tfx/examples/cifar10/cifar10_utils_native_keras.py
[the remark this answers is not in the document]
davidzats-eng: The evaluator should work out-of-the-box with both serialized examples as well as tensor-based inputs. What error do you see?

tfx/examples/cifar10/meta_data_writer.py
Deanplayerljx: could you elaborate this a little bit? I'm not very sure what wrap in parenthesis mean..

tfx/examples/cifar10/cifar10_utils_native_keras.py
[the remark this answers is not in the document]
davidzats-eng: sounds good. as long as it is more clear.

tfx/examples/cifar10/cifar10_utils_native_keras.py
Deanplayerljx: yes this is never used, I thought it might be strange to return an uncomplied model in "build_keras_model" function :p Also this makes it more consistent with MNIST example. If you feel it's unnecessary i can remove it. what do you think?

tfx/examples/cifar10/cifar10_utils_native_keras.py
Deanplayerljx: I think model.trainable basically set all layers' "trainable" attribute to true, which is equal to _freeze_model_by_percentage(0.0).

tfx/examples/cifar10/cifar10_utils_native_keras.py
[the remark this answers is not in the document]
davidzats-eng: sounds good, lets find this empirically.

tfx/examples/cifar10/cifar10_utils_native_keras.py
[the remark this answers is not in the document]
davidzats-eng: great! thanks for checking. I completely agree that if additional layers don't help we should not be adding them.

tfx/examples/cifar10/cifar10_utils_native_keras.py
davidzats-eng: I would rather not confuse people with unnecessary computation, so lets go ahead and remove. Also, see if you can make the code a bit more readable with a few helper functions.

tfx/examples/cifar10/meta_data_writer.py
davidzats-eng: sure, we commonly avoid using "\" for newlines and instead use parenthesis to split lines.

tfx/examples/cifar10/meta_data_writer.py
davidzats-eng: how do you feel about having a commandline flag of the form: also_produce_json_metadata?

tfx/examples/cifar10/meta_data_writer.py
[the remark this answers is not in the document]
davidzats-eng: sure, my point here is that the model file has to be named cifar10.tflite for the appropriate model_info to be used. Instead we could allow any model file and pass cifar10.tflite as a commandline argument.

tfx/examples/cifar10/cifar10_utils_native_keras.py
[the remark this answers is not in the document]
Deanplayerljx: I mean it's not the problem with evaluator, and the evaluator will work in both case as long as we feed it with the correct input, i.e. transformed/not transformed. I think the problem is with the saved TFLite model, which will be used for serving in other frameworks. If I understand the process correctly, if we include the transform layer into the model, the saved TFLite model will need to take encoded image strings as input instead of image tensors, which is not compatible with serving frameworks like MLKit and MediaPipe.

tfx/examples/cifar10/README.md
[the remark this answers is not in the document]
davidzats-eng: see below, given the assumptions of your metadata_writer, you must currently rename it. I agree it is more informative, but from an example / documentation standpoint it may be simpler to just go with tflite.

tfx/examples/cifar10/meta_data_writer.py
Deanplayerljx: sure sounds good

tfx/examples/cifar10/README.md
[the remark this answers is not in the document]
Deanplayerljx: sure sounds good

tfx/examples/cifar10/meta_data_writer.py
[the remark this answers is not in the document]
Deanplayerljx: got you

tfx/examples/cifar10/meta_data_writer.py
Deanplayerljx: sure sounds good

tfx/examples/cifar10/cifar10_utils_native_keras.py
Deanplayerljx: sounds good

tfx/examples/cifar10/meta_data_writer.py
Deanplayerljx: i see, will change it

tfx/examples/cifar10/cifar10_utils_native_keras.py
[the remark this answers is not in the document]
Deanplayerljx: 4 labels in cifar-10 do not exist in ImageNet, and 6 labels "some what" exist. The labels in ImageNet are fine-grained categories, for example, there are 5 labels in ImageNet that are related to cat: "tabby cat", "tiger cat", "Persian cat", "Siamese cat", "Egyptian cat", whereas in cifar-10 there is only one label named "cat".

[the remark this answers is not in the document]
davidzats-eng: ok cool, can we check how our model does on those cifar-10 labels?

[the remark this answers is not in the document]
Deanplayerljx: hmmm i feel there is no direct way to do this? The origin pre-trained model has Dense(1000) as the final classifier, and each output corresponds to one ImageNet class. Since there is no one-to-one mapping between these 1000 classes and cifar-10 class, i think we could not test its performance for cifar-10?

Deanplayerljx:
@googlebot I fixed it.

Deanplayerljx:
@googlebot I consent

tfx/examples/cifar10/cifar10_utils_native_keras.py
Deanplayerljx: model.trainable is like a meta switch that turns all parameters trainable or not. if we freeze some layers in the model and then call model.trainable=True, all layers will be unfrozen; if we call model.trainable=False after freezing some of layers, all layers will be frozen.

davidzats-eng: Makes sense. The question is what is the default and does it get in the way.

Deanplayerljx: The default for model.trainable is True. As long as we don't modify it it will not get in the way.

Deanplayerljx:
It seems I messed up this PR with bunch of other people's commits. will open a new clean PR for this

PR #2098, opened 2020-07-06
Support Custom Split Config for TFDV Related Components

tinally:
# Support custom split config for TFDV related components by adding exclude_splits param to StatisticsGen, SchemaGen and ExampleValidator.


tfx/components/example_validator/component.py
```diff
@@ -95,6 +99,9 @@ def __init__(self,
         type=standard_artifacts.ExampleAnomalies,
         artifacts=[standard_artifacts.ExampleAnomalies()])
     spec = ExampleValidatorSpec(
-        statistics=statistics, schema=schema, anomalies=anomalies)
+        statistics=statistics,
+        schema=schema,
+        exclude_splits=exclude_splits,
+        anomalies=anomalies)
```
1025KB, line 105: you need to update executor too, it's currently hardcode a split I think, same for schema_gen executor

tfx/components/example_validator/component.py
```diff
@@ -68,6 +68,8 @@ class ExampleValidator(base_component.BaseComponent):
   def __init__(self,
                statistics: types.Channel = None,
                schema: types.Channel = None,
+               exclude_splits: Optional[List[Union[
+                   Text, data_types.RuntimeParameter]]] = None,
```
1025KB, line 72: could you double check if this data_types.RuntimeParameter works, can ExampleValidatorSpec parse it automatically?

tfx/components/example_validator/component.py
```diff
@@ -78,6 +80,8 @@ def __init__(self,
         should contain at least 'eval' split. Other splits are currently
         ignored.
       schema: A Channel of type `standard_artifacts.Schema`. _required_
+      exclude_splits: Names of splits that the example validator should not
+        validate.
```
1025KB, line 84: I think we need a default value for it, so existing user don't need to set it to have the current behavior

tfx/components/example_validator/component.py
tinally: yes i am updating executor, i will change the hardcode

tfx/components/statistics_gen/executor.py
```diff
@@ -107,10 +109,12 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
         stats_options.schema = schema
 
     split_uris = []
+    exclude_splits = exec_properties['exclude_splits']
```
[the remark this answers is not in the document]
tinally: got it, i changed it

tfx/components/example_validator/component.py
tinally: i removed RuntimeParameter for now.

tfx/components/example_validator/component_test.py
```diff
@@ -31,10 +31,21 @@ def testConstruct(self):
...
+  def testConstructWithExcludeSplits(self):
+    statistics_artifact = standard_artifacts.ExampleStatistics()
+    statistics_artifact.split_names = artifact_utils.encode_split_names(
+        ['train', 'eval'])
+    exclude_splits = ['train']
     example_validator = component.ExampleValidator(
         statistics=channel_utils.as_channel([statistics_artifact]),
         schema=channel_utils.as_channel([standard_artifacts.Schema()]),
-    )
+        exclude_splits=exclude_splits)
```
1025KB, line 48: you can just add it to a existing test, it's trivial change

tfx/components/schema_gen/component_test.py
```diff
@@ -52,6 +52,18 @@ def testConstructWithParameter(self):
         str(schema_gen.spec.exec_properties['infer_feature_shape']),
         str(infer_shape))
 
+  def testConstructWithExcludeSplits(self):
```
1025KB, line 55: same here, just add it to testConstruct is fine

tfx/components/example_validator/component_test.py
tinally: got it, i updated this

tfx/components/example_validator/component.py
tinally: yes, added the default behaviors

tfx/components/example_validator/component.py
```diff
@@ -68,6 +69,8 @@ class ExampleValidator(base_component.BaseComponent):
   def __init__(self,
                statistics: types.Channel = None,
                schema: types.Channel = None,
+               # pylint: disable=dangerous-default-value
+               exclude_splits: Optional[List[Text]] = ['train'],
```
1025KB, line 73: if we don't allow None, we can set it to, then it's not a dangerous default?

exclude_splits:List[Text] = ['train'] (user can set it to [] if they don't want to exclude)

same elsewhere

tfx/components/example_validator/component.py
```diff
@@ -91,10 +96,17 @@ def __init__(self,
           'been renamed to "statistics" and is deprecated. Please update your '
           'usage as support for this argument will be removed soon.')
       statistics = stats
+    if not output:
+      anomalies_artifact = standard_artifacts.ExampleAnomalies()
+      anomalies_artifact.split_names = artifact_utils.get_single_instance(
+          list(statistics.get())).split_names
     anomalies = output or types.Channel(
```
1025KB, line 103: move this within the if block

tfx/components/example_validator/component_test.py
```diff
@@ -30,11 +30,12 @@ class ExampleValidatorTest(tf.test.TestCase):
...
     statistics_artifact = standard_artifacts.ExampleStatistics()
     statistics_artifact.split_names = artifact_utils.encode_split_names(
-        ['eval'])
+        ['train', 'eval'])
+    exclude_splits = ['train']
     example_validator = component.ExampleValidator(
         statistics=channel_utils.as_channel([statistics_artifact]),
         schema=channel_utils.as_channel([standard_artifacts.Schema()]),
-    )
+        exclude_splits=exclude_splits)
```
1025KB, line 38: check if example_validator.exec_properties contains the correct value, same elsewhere

tfx/components/example_validator/executor.py
```diff
@@ -63,30 +66,40 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
-            io_utils.SchemaReader().read(
-                io_utils.get_only_uri_in_dir(
-                    artifact_utils.get_single_uri(input_dict[SCHEMA_KEY])))
-    }
-    output_uri = artifact_utils.get_single_uri(output_dict[ANOMALIES_KEY])
-    label_outputs = {labels.SCHEMA_DIFF_PATH: output_uri}
-    self._Validate(label_inputs, label_outputs)
-    absl.logging.info(
-        'Validation complete. Anomalies written to {}.'.format(output_uri))
+    exclude_splits = exec_properties.get(EXCLUDE_SPLITS_KEY)
```
1025KB, line 78: you can just use exec_properties['EXCLUDE_SPLITS_KEY'] as EXCLUDE_SPLITS_KEY will also be in the exec_properties map, just the value might be None (you have not exclude_splits to cover that below)

tfx/components/example_validator/executor.py
```diff
@@ -63,30 +66,40 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+    schema = io_utils.SchemaReader().read(
+        io_utils.get_only_uri_in_dir(
+            artifact_utils.get_single_uri(input_dict[SCHEMA_KEY])))
+    for artifact in input_dict[STATISTICS_KEY]:
+      for split in artifact_utils.decode_split_names(artifact.split_names):
+        if not exclude_splits or split not in exclude_splits:
```
1025KB, line 84: do this 

```


for xxx
   if exclude_splits and split in exclude_splits:
     continuous
   process...
```

same elsewhere

tfx/components/example_validator/executor.py
```diff
@@ -63,30 +66,40 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+          absl.logging.info(
+              'Validating schema against the computed statistics.')
```
1025KB, line 86: log the split name here

tfx/components/example_validator/executor.py
```diff
@@ -63,30 +66,40 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+              labels.SCHEMA:
+                  schema
+          }
+          output_uri = artifact_utils.get_split_uri(output_dict[ANOMALIES_KEY],
+                                                    split)
+          label_outputs = {labels.SCHEMA_DIFF_PATH: output_uri}
+          self._Validate(label_inputs, label_outputs)
+          absl.logging.info(
+              'Validation complete. Anomalies written to {}.'.format(
+                  output_uri))
```
1025KB, line 102: log split name

tfx/components/example_validator/component.py
```diff
@@ -91,10 +96,17 @@ def __init__(self,
           'been renamed to "statistics" and is deprecated. Please update your '
           'usage as support for this argument will be removed soon.')
       statistics = stats
+    if not output:
+      anomalies_artifact = standard_artifacts.ExampleAnomalies()
+      anomalies_artifact.split_names = artifact_utils.get_single_instance(
+          list(statistics.get())).split_names
```
1025KB, line 102: I think you need to exclude split in the split_names too, same for stats_gen

tfx/components/example_validator/executor_test.py
```diff
@@ -46,26 +46,82 @@ def testDo(self):
...
+    self.assertEqual(['anomalies.pbtxt'],
+                     tf.io.gfile.listdir(os.path.join(validation_output.uri,
+                                                      'eval')))
+    anomalies = io_utils.parse_pbtxt_file(
+        os.path.join(validation_output.uri, 'eval', 'anomalies.pbtxt'),
+        anomalies_pb2.Anomalies())
+    self.assertNotEqual(0, len(anomalies.anomaly_info))
+    # TODO(zhitaoli): Add comparison to expected anomolies.
+
+  def testDoWithExcludedSplits(self):
```
1025KB, line 77: you can just add it to testDo, it's not two different branch in the code as ExcludedSplit is in your main code branch

tfx/components/schema_gen/executor_test.py
```diff
@@ -46,11 +46,48 @@ def testDo(self):
...
     output_dict = {
         executor.SCHEMA_KEY: [schema_output],
     }
 
-    exec_properties = {'infer_feature_shape': False}
+    schema_gen_executor = executor.Executor()
+    schema_gen_executor.Do(input_dict, output_dict, exec_properties)
+    self.assertNotEqual(0, len(tf.io.gfile.listdir(schema_output.uri)))
+
+  def testDoWithExcludeSplits(self):
```
1025KB, line 63: merge test as comments above

and update_schema is not tested, consider create a fake third split in testdata (e.g., `test` that copy from `eval`) thus you can test both infer and update schema

tfx/components/statistics_gen/component_test.py
```diff
@@ -55,6 +55,16 @@ def testConstructWithSchemaAndStatsOptions(self):
     self.assertEqual(standard_artifacts.ExampleStatistics.TYPE_NAME,
                      statistics_gen.outputs['statistics'].type_name)
 
+  def testConstructWithExcludeSplits(self):
```
1025KB, line 58: just add to testConstruct like other component test

tfx/components/statistics_gen/executor.py
```diff
@@ -38,6 +38,7 @@
 
 # Keys for exec_properties dict.
 STATS_OPTIONS_JSON_KEY = 'stats_options_json'
+EXCLUDE_SPLITS_KEY = 'exclude_splits'
```
1025KB, line 41: be constant with format (e.g., comments)

tfx/components/statistics_gen/executor.py
```diff
@@ -107,10 +109,12 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
         stats_options.schema = schema
 
     split_uris = []
+    exclude_splits = exec_properties[EXCLUDE_SPLITS_KEY]
     for artifact in input_dict[EXAMPLES_KEY]:
       for split in artifact_utils.decode_split_names(artifact.split_names):
-        uri = os.path.join(artifact.uri, split)
-        split_uris.append((split, uri))
+        if not exclude_splits or split not in exclude_splits:
+          uri = os.path.join(artifact.uri, split)
```
1025KB, line 116: let's use the artifact_utils.get_split_uri(xxx, split) (this should check if split_name is set correctly)

tfx/components/statistics_gen/executor_test.py
```diff
@@ -164,8 +171,49 @@ def testDoWithTwoSchemas(self):
     # Run executor.
     stats_gen_executor = executor.Executor()
     with self.assertRaises(ValueError):
-      stats_gen_executor.Do(
-          input_dict, output_dict, exec_properties=exec_properties)
+      stats_gen_executor.Do(input_dict, output_dict, exec_properties)
+
+  def testDoWithExcludeSplits(self):
```
1025KB, line 176: consider merge the test (you can create a fake split, and exclude that, so you test both `exclude` functionality and multiple splits output functionality)

tfx/components/example_validator/component.py
tinally: then should i change the parameters in the spec so optional is set to False? I will update design doc as well?

tinally: I kept the pylint issue for now and let's see what other reviewers prefer.

tfx/components/schema_gen/executor_test.py
tinally: i tested update_schema and added log statements here as well.

tfx/components/example_validator/component.py
1025KB: sgtm, 
(discussed offline, the other option is 
use None as default (will be changed to ['train']) and use [] as no exclude splits, in this option, none and empty means different things.)

tfx/components/example_validator/component.py
```diff
@@ -78,6 +81,10 @@ def __init__(self,
         should contain at least 'eval' split. Other splits are currently
         ignored.
       schema: A Channel of type `standard_artifacts.Schema`. _required_
+      exclude_splits: Names of splits that the example validator should not
+        validate. If exclude_splits is None or an empty list, no splits will
+        be excluded. Default behavior is validating the statistics of 'eval'
+        splits.
```
1025KB, line 87: default behavior is excluding the `train` split

tfx/components/example_validator/component.py
```diff
@@ -91,10 +98,22 @@ def __init__(self,
...
-        type=standard_artifacts.ExampleAnomalies,
-        artifacts=[standard_artifacts.ExampleAnomalies()])
+    anomalies = output
+    if not output:
+      anomalies_artifact = standard_artifacts.ExampleAnomalies()
+      split_names = artifact_utils.get_single_instance(
+          list(statistics.get())).split_names
+      for split in split_names:
+        if exclude_splits and split in exclude_splits:
+          split_names.remove(split)
```
1025KB, line 108: there is a short way for this 

[i for i in alist if i not in blist]

https://www.geeksforgeeks.org/python-remove-all-values-from-a-list-present-in-other-list/

tfx/components/example_validator/executor.py
```diff
@@ -63,30 +66,40 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+    exclude_splits = exec_properties[EXCLUDE_SPLITS_KEY]
+    schema = io_utils.SchemaReader().read(
+        io_utils.get_only_uri_in_dir(
+            artifact_utils.get_single_uri(input_dict[SCHEMA_KEY])))
+    for artifact in input_dict[STATISTICS_KEY]:
+      for split in artifact_utils.decode_split_names(artifact.split_names):
+        if exclude_splits and split in exclude_splits:
+          continue
+        absl.logging.info('Validating schema against the computed statistics '
+                          'for split {}.'.format(split))
```
1025KB, line 87: we prefer [this](https://github.com/tensorflow/tfx/blob/master/tfx/orchestration/beam/beam_dag_runner.py#L165) style now

same elsewhere

tfx/components/example_validator/executor_test.py
```diff
@@ -46,26 +46,45 @@ def testDo(self):
         self._testMethodName)
 
     validation_output = standard_artifacts.ExampleAnomalies()
-    validation_output.uri = os.path.join(output_data_dir, 'output')
+    validation_output.uri = output_data_dir
```
1025KB, line 49: let's just use os.path.join(output_data_dir, 'output'), thus if we have other output in the future, it's not conflict

tfx/components/example_validator/component.py
zhitaoli: Don't use list for default value because it's **mutable**

https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

Use None here and make it clear that this would default to ['train'] if not set

tfx/components/example_validator/executor_test.py
```diff
@@ -46,26 +46,45 @@ def testDo(self):
...
+                                                      'train')))
+    train_anomalies = io_utils.parse_pbtxt_file(
+        os.path.join(validation_output.uri, 'train', 'anomalies.pbtxt'),
         anomalies_pb2.Anomalies())
-    self.assertNotEqual(0, len(anomalies.anomaly_info))
+    self.assertEqual(0, len(train_anomalies.anomaly_info))
+
+    self.assertEqual(['anomalies.pbtxt'],
+                     tf.io.gfile.listdir(os.path.join(validation_output.uri,
+                                                      'eval')))
```
1025KB, line 78: how about this

```
eval_anomalies_path = os.path.join(validation_output.uri, 'eval', 'anomalies.pbtxt')
self.assertTrue(tf.io.gfile.exists(eval_anomalies_path))
eval_anomalies = io_utils.parse_pbtxt_file(eval_anomalies_path, ...)
self.assertNotEqual(0, len(eval_anomalies.anomaly_info))

```

tfx/components/schema_gen/component.py
```diff
@@ -75,6 +77,10 @@ def __init__(
         whether or not to infer the shape of features. If the feature shape is
         not inferred, downstream Tensorflow Transform component using the schema
         will parse input as tf.SparseTensor.
+      exclude_splits: Names of splits that will not be taken into consideration
+        when auto-generating a schema. If exclude_splits is None or an empty
+        list, no splits will be excluded. Default behavior is generating a
+        schema based on 'eval' splits.
```
1025KB, line 83: exclude eval

tfx/components/schema_gen/executor.py
```diff
@@ -59,21 +61,37 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
-        artifact_utils.get_split_uri(input_dict[STATISTICS_KEY], 'train'))
+    infer_feature_shape = exec_properties['infer_feature_shape']
+    exclude_splits = exec_properties[EXCLUDE_SPLITS_KEY]
+    # Only one schema is generated for all splits.
+    schema = None
+    for artifact in input_dict[STATISTICS_KEY]:
+      for split in artifact_utils.decode_split_names(artifact.split_names):
+        if exclude_splits and split in exclude_splits:
+          continue
+        stats_uri = io_utils.get_only_uri_in_dir(
```
1025KB, line 80: add a empty line after `continue`, same else where

tfx/components/schema_gen/executor.py
```diff
@@ -59,21 +61,37 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+            artifact_utils.get_split_uri(input_dict[STATISTICS_KEY], split))
+        if not schema:
+          absl.logging.info('Infering schema from statistics for split '
+                            '{}.'.format(split))
+          schema = tfdv.infer_schema(
+              tfdv.load_statistics(stats_uri), infer_feature_shape)
+        else:
+          absl.logging.info('Updating schema from statistics for split '
```
1025KB, line 88: maybe just merge the logging to 'process schema for split xxx'

as we don't care which one use infer which one use update, the order doesn't matter, we just need to know all required splits are processed

tfx/components/example_validator/component.py
tinally: yes, i saw this, but exclude_lists might be none so we need to check for that as well so it will be [i for i in alist if not blist or i not in blist]. 

with formatting, this will also take three lines, and it's looks more complicated.

should i still use the short way?

tfx/components/example_validator/component.py
tinally: Thanks. I understand. I will make the change now.

tfx/components/example_validator/executor.py
tinally: got it, i changed all the logging statements

tfx/components/example_validator/component.py
```diff
@@ -91,10 +97,25 @@ def __init__(self,
           'been renamed to "statistics" and is deprecated. Please update your '
           'usage as support for this argument will be removed soon.')
       statistics = stats
-    anomalies = output or types.Channel(
-        type=standard_artifacts.ExampleAnomalies,
-        artifacts=[standard_artifacts.ExampleAnomalies()])
+    if exclude_splits is None:
+      exclude_splits = ['train']
```
1025KB, line 101: print an info log like "Exclude `train` split when exclude_splits is unset (None)"

tfx/components/example_validator/component.py
```diff
@@ -91,10 +97,25 @@ def __init__(self,
...
+    anomalies = output
+    if not output:
```
1025KB, line 103: if not anomalies

tfx/components/example_validator/executor.py
```diff
@@ -63,30 +66,41 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
-    self._Validate(label_inputs, label_outputs)
-    absl.logging.info(
-        'Validation complete. Anomalies written to {}.'.format(output_uri))
+    exclude_splits = exec_properties[EXCLUDE_SPLITS_KEY]
+    schema = io_utils.SchemaReader().read(
+        io_utils.get_only_uri_in_dir(
+            artifact_utils.get_single_uri(input_dict[SCHEMA_KEY])))
+    for artifact in input_dict[STATISTICS_KEY]:
+      for split in artifact_utils.decode_split_names(artifact.split_names):
+        if exclude_splits and split in exclude_splits:
```
1025KB, line 84: exclude_splits won't be None now, but keep this exclude_splits is no harm, up to you

tfx/components/statistics_gen/component.py
```diff
@@ -86,8 +90,12 @@ def __init__(self,
       examples = input_data
     if not output:
       statistics_artifact = standard_artifacts.ExampleStatistics()
-      statistics_artifact.split_names = artifact_utils.get_single_instance(
+      split_names = artifact_utils.get_single_instance(
           list(examples.get())).split_names
+      for split in split_names:
+        if exclude_splits and split in exclude_splits:
+          split_names.remove(split)
```
1025KB, line 97: same as above

btw, is split_names a string? if so how can remove it work?

tfx/components/statistics_gen/executor_test.py
```diff
@@ -105,6 +115,8 @@ def testDoWithSchemaAndStatsOptions(self):
     exec_properties = {
         executor.STATS_OPTIONS_JSON_KEY:
             tfdv.StatsOptions(label_feature='company').to_json(),
+        executor.EXCLUDE_SPLITS_KEY:
+            None
```
1025KB, line 119: it won't be None for executor (as component will set it to default), same elsewhere

tinally: yes, to be consistent with schema gen and example validator, i think we also remove the case for exclude_splits = None for statsgen

tfx/components/example_validator/executor.py
tinally: i think this is good convention to confirm the list we are accessing is not none? if this is too extra, i can take this out

tfx/components/statistics_gen/component_test.py
```diff
@@ -31,10 +31,17 @@ class ComponentTest(tf.test.TestCase):
...
+    exclude_splits = ['eval']
     statistics_gen = component.StatisticsGen(
-        examples=channel_utils.as_channel([examples]))
+        examples=channel_utils.as_channel([examples]),
+        exclude_splits=exclude_splits)
     self.assertEqual(standard_artifacts.ExampleStatistics.TYPE_NAME,
                      statistics_gen.outputs['statistics'].type_name)
+    self.assertEqual(['train'], artifact_utils.decode_split_names(
+        artifact_utils.get_single_instance(
+            list(statistics_gen.outputs['statistics'].get())).split_names))
```
1025KB, line 42: you can just test the split_name string, no need to decode (same elsewhere)

tfx/components/example_validator/component.py
```diff
@@ -91,10 +98,27 @@ def __init__(self,
           'been renamed to "statistics" and is deprecated. Please update your '
           'usage as support for this argument will be removed soon.')
       statistics = stats
-    anomalies = output or types.Channel(
-        type=standard_artifacts.ExampleAnomalies,
-        artifacts=[standard_artifacts.ExampleAnomalies()])
+    if exclude_splits is None:
+      exclude_splits = ['train']
```
brills, line 102: why is the train split always excluded from validation?

tfx/components/statistics_gen/executor.py
```diff
@@ -89,12 +94,11 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
-    if STATS_OPTIONS_JSON_KEY in exec_properties:
+    if exec_properties.get(STATS_OPTIONS_JSON_KEY):
       stats_options_json = exec_properties[STATS_OPTIONS_JSON_KEY]
-      if stats_options_json:
-        # TODO(b/150802589): Move jsonable interface to tfx_bsl and use
-        # json_utils
-        stats_options = options.StatsOptions.from_json(stats_options_json)
+      # TODO(b/150802589): Move jsonable interface to tfx_bsl and use
+      # json_utils
+      stats_options = options.StatsOptions.from_json(stats_options_json)
```
brills, line 101: since you changed it.. make it even concise and avoid double lookups?

```
stats_options_json = exec_properties.get(STATS_OPTIONS_JSON_KEY)
if stats_options_json is not None:
  ...
``

File: tfx/components/statistics_gen/executor.py [hunk 37]
```diff
@@ -107,26 +111,30 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
         stats_options.schema = schema
 
     split_uris = []
+    exclude_splits = exec_properties[EXCLUDE_SPLITS_KEY]
```
brills, line 114: use `.get()` instead to avoid mutating exec_properties.

tfx/components/schema_gen/component.py
```diff
@@ -88,11 +93,15 @@ def __init__(
           'been renamed to "statistics" and is deprecated. Please update your '
           'usage as support for this argument will be removed soon.')
       statistics = stats
+    if exclude_splits is None:
```
brills, line 96: again, why exclude the eval dataset?

Note that the schema is critical in guiding the parsing of features. We've seen examples where feature "foo" is present in all the training examples, thus the generated schema says it should be parsed into a dense Tensor, however the same feature is missing in the eval examples, leading to parsing failures.

tfx/components/example_validator/component.py
tinally: because this is the current behavior -- before we added exclude_splits, example validator is hardcode to only use 'eval' split (in other words, 'train' split is excluded). you can see the 'current behavior' listed for each component in the design doc.

tfx/components/statistics_gen/executor.py
tinally: yes, i think this is more concise, thanks for the catch

tfx/components/schema_gen/component.py
tinally: currently, the hardcode behavior is generating a schema based on ‘train’ split. therefore, when the user doesn't provide exclude_splits, we want to keep it as its current behavior (i.e. exclude 'eval' and only use 'train'). 

for this parsing failure you mentioned, are there any solution? how can we avoid this parsing failure if a feature is only present in all training examples but not 'eval' examples?

tfx/components/statistics_gen/executor.py
tinally: we actually defined exclude_splits to be an empty list if it's not provided so i can simplify line 117 directly.

tfx/components/example_validator/component.py
```diff
@@ -78,6 +80,10 @@ def __init__(self,
         should contain at least 'eval' split. Other splits are currently
         ignored.
       schema: A Channel of type `standard_artifacts.Schema`. _required_
+      exclude_splits: Names of splits that the example validator should not
+        validate. Default behavior (when exclude_splits is set to None)
+        is excluding no splits.
+      output: Output `Schema` channel for schema result.
```
1025KB, line 86: remove this `output`

tfx/components/example_validator/executor_test.py
```diff
@@ -47,25 +47,43 @@ def testDo(self):
...
+    eval_anomalies_path = os.path.join(validation_output.uri, 'eval',
+                                       'anomalies.pbtxt')
+    self.assertTrue(tf.io.gfile.exists(train_anomalies_path))
+    self.assertTrue(tf.io.gfile.exists(eval_anomalies_path))
+    train_anomalies = io_utils.parse_pbtxt_file(train_anomalies_path,
+                                                anomalies_pb2.Anomalies())
+    eval_anomalies = io_utils.parse_pbtxt_file(eval_anomalies_path,
+                                               anomalies_pb2.Anomalies())
+    self.assertEqual(0, len(train_anomalies.anomaly_info))
+    self.assertNotEqual(0, len(eval_anomalies.anomaly_info))
```
1025KB, line 81: schema under testdata might need to be updated, current schema is generated based on train, thus there isn't anomaly_info for train. since we change it to both train and eval, there shouldn't be anomaly_info for eval too

tfx/components/statistics_gen/executor.py
```diff
@@ -89,12 +94,11 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
 
     stats_options = options.StatsOptions()
-    if STATS_OPTIONS_JSON_KEY in exec_properties:
-      stats_options_json = exec_properties[STATS_OPTIONS_JSON_KEY]
-      if stats_options_json:
-        # TODO(b/150802589): Move jsonable interface to tfx_bsl and use
-        # json_utils
-        stats_options = options.StatsOptions.from_json(stats_options_json)
+    stats_options_json = exec_properties.get(STATS_OPTIONS_JSON_KEY)
+    if stats_options_json is not None:
```
1025KB, line 98: if stats_options_json:

tfx/types/standard_component_specs.py
```diff
@@ -100,7 +100,9 @@ class EvaluatorSpec(ComponentSpec):
 class ExampleValidatorSpec(ComponentSpec):
   """ExampleValidator component spec."""
 
-  PARAMETERS = {}
+  PARAMETERS = {
+      'exclude_splits': ExecutionParameter(type=List[Text], optional=True),
```
1025KB, line 104: it might need to be List[(str, Text)] or (List[str], List[Text]), please check

tfx/components/example_validator/component_test.py
```diff
@@ -30,13 +30,18 @@ class ExampleValidatorTest(tf.test.TestCase):
...
     self.assertEqual(standard_artifacts.ExampleAnomalies.TYPE_NAME,
                      example_validator.outputs['anomalies'].type_name)
+    self.assertEqual('["train"]', artifact_utils.get_single_instance(
+        list(example_validator.outputs['anomalies'].get())).split_names)
+    self.assertEqual(example_validator.spec.exec_properties['exclude_splits'],
+                     ['eval'])
```
1025KB, line 44: exec_properties['exclude_splits'] doesn't allow list type, we might need to convert it to json str

tfx/components/example_validator/component.py
tinally: ah, my mistake, thanks

tfx/types/standard_component_specs.py
tinally: both of these will fail the component tests

tfx/components/statistics_gen/executor.py
```diff
@@ -106,27 +111,37 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
                 artifact_utils.get_single_uri(input_dict[SCHEMA_KEY])))
         stats_options.schema = schema
 
+    # Load and deserialize exclude splits from execution properties.
+    exclude_splits = json_utils.loads(
+        exec_properties.get(EXCLUDE_SPLITS_KEY)) or []
```
1025KB, line 116: json_utils.loads(exec_properties.get(EXCLUDE_SPLITS_KEY, 'null')) or [], same elsewhere

check my comments on the evaluator PR

tfx/components/example_validator/component.py
```diff
@@ -17,16 +17,18 @@
 from __future__ import division
 from __future__ import print_function
 
-from typing import Optional, Text
+from typing import Optional, List, Text
```
1025KB, line 20: did you try if dsl compiler pass or not?

tinally: it works now, dsl compiler test is passed.

tfx/components/example_gen/csv_example_gen/executor.py
```diff
@@ -31,6 +31,24 @@
...
+    value_list.append(int(cell))
+  return tf.train.Feature(int64_list=tf.train.Int64List(value=value_list))
+
+def _float_handler(cell):
+  value_list = []
+  if cell:
+    value_list.append(float(cell))
+  return tf.train.Feature(float_list=tf.train.FloatList(value=value_list))
+
+def _bytes_handler(cell):
```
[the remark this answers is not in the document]
tinally: ah my mistake, i thought you meant adding beam typehints, it's fixed now

tfx/components/example_gen/csv_example_gen/executor.py
```diff
@@ -31,35 +31,52 @@
 from tfx.utils import io_utils
 
 
+def _int_handler(cell: bytes) -> tf.train.Feature:
```
1025KB, line 34: is bytes or csv_decoder.CSVCell? check line 93 & 99

tfx/components/example_gen/csv_example_gen/executor.py
```diff
@@ -31,35 +31,52 @@
...
-            bytes_list=tf.train.BytesList(value=[csv_cell]))
+        handler_fn = _bytes_handler
       else:
         handler_fn = None
       column_handlers.append((column_info.name, handler_fn))
 
     self._column_handlers = column_handlers
 
   def process(
+      #pylint: disable=arguments-differ
```
1025KB, line 79: this will disable this lint for the entire file, not for the single line

it should be added at end of line 80 or 81 depends on which line has the lint issue

tinally: i tried putting it at the end (after ':') but it still has the pylint error, let me try putting it at end of line 80

tfx/components/example_gen/csv_example_gen/executor.py
tinally: i printed out the type of cell inside the method and it was 'bytes', if it's type csv_decoder.CSVCell, it should print out that?

1025KB: check tfx_bsl/coders/csv_decoder.py 
csv_decoder.CSVCell = bytes,

let use either cell:CSVCell or value:bytes

tinally: i see, i used cell: CSVCell

tfx/components/example_gen/csv_example_gen/executor.py
tinally: i searched and i think because the pylint issue was said to be on line 78, so the disable message can be added to line 78 as well

PR #2124, opened 2020-07-10
TFX on Beam Runner Benchmarking

kshivvy:
Added functionality for running the existing TFMA, TFMA V2, and TFT tests that used Apache Beam pipelines to run on local scaled execution, Cloud Dataflow, and the flink-on-k8s-operator.


kshivvy:
@charlesccychen

tfx/benchmarks/beam_pipeline_benchmark_chicago_taxi.py
```diff
@@ -0,0 +1,38 @@
...
+
+from tfx.benchmarks import beam_pipeline_benchmark_base
+
+
+class BeamPipelineBenchmarkChicagoTaxi(
+    beam_pipeline_benchmark_base.BeamPipelineBenchmarkBase):
+
+  def __init__(self, **kwargs):
+    super(BeamPipelineBenchmarkChicagoTaxi, self).__init__(
+        min_num_workers=3, max_num_workers=4,
```
kshivvy, line 28: Shuld be from 1 to 32

tfx/benchmarks/beam_pipeline_benchmark_chicago_taxi.py
```diff
@@ -0,0 +1,38 @@
...
+        base_dir="gs://tfx-keshav-example-bucket/datasets",
```
kshivvy, line 29: What should I change this base_dir to - is there a generic bucket I can host this on?

tfx/benchmarks/benchmark_base.py
```diff
@@ -27,14 +28,73 @@
...
+
+def set_beam_pipeline_mode(mode):
+  assert mode in modes
+  globals()['beam_pipeline_mode'] = mode
+
+def set_num_workers(num_workers):
+  globals()['num_workers_str'] = str(num_workers)
+
+def set_cloud_dataflow_temp_loc(temp_loc):
+  globals()['cloud_dataflow_temp_loc'] = temp_loc
```
yifanmai, line 44: Could we make these instance variables instead of globals?

You'd also have to make these methods instead of global functions, and possibly initialize the instance variables to `None` in the `__init__`. Alternatively, you could subclass `BenchmarkBase` to a new subclass to preserve the backwards compatibility of `BenchmarkBase`.

tfx/benchmarks/benchmark_base.py
```diff
@@ -27,14 +28,73 @@
     "Beam runner to use - any runner name accepted by "
     "apache_beam.runners.create_runner")
 
+modes = ["default", "local_scaled_execution", "cloud_dataflow", "flink_on_k8s"]
```
yifanmai, line 31: nit: use `ALL_CAPS` for constants as per [PEP-8](https://www.python.org/dev/peps/pep-0008/#constants).

other nit: define these modes separately first
```
DEFAULT_MODE = "default"
LOCAL_SCALED_EXECUTION_MODE = "local_scaled_execution"
modes = [DEFAULT_MODE, LOCAL_SCALED_EXECUTION, ...]
```
then you can used the named constants later.

tfx/benchmarks/benchmark_base.py
```diff
@@ -27,14 +28,73 @@
...
     # test framework (e.g. PerfZero creates the class and calls the methods
     # directly)
     runner_flag = (
         FLAGS.beam_runner
         if FLAGS.is_parsed() else FLAGS["beam_runner"].default)
     return beam.Pipeline(runner=beam.runners.create_runner(runner_flag))
+
+  def _create_beam_pipeline(self):
+
+    if beam_pipeline_mode == "default":
```
yifanmai, line 87: nit: used named constants (see earlier)

tfx/benchmarks/benchmark_base.py
```diff
@@ -27,14 +28,73 @@
...
+
+def set_cloud_dataflow_temp_loc(temp_loc):
+  globals()['cloud_dataflow_temp_loc'] = temp_loc
 
 class BenchmarkBase(test.Benchmark):
+  """Base class for running Beam pipelines on various runners"""
 
-  def _create_beam_pipeline(self):
+  def _set_cloud_dataflow_flags(self):
+    self.flags = ['--runner=DataflowRunner',
```
yifanmai, line 50: nit: in general we prefer the same type (single or double) of quotes for all strings within a file, even though [PEP-8](https://legacy.python.org/dev/peps/pep-0008/#string-quotes) doesn't make an explicit recommendation about it.

tfx/benchmarks/beam_pipeline_benchmark_base.py
```diff
@@ -0,0 +1,198 @@
...
+      self.wall_times_list.append(test_names)
+
+    # Add wall times to dataset
+    row = [self.num_workers]
+    for tf_module in self.wall_times:
+      for test, wall_time in self.wall_times[tf_module].items():
+        row.append(wall_time)
+
+    self.wall_times_list.append(row)
+    print(self.wall_times_list)
```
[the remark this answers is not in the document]
kshivvy: Nope, was just there to help with debugging. I'll remove it

tfx/benchmarks/beam_pipeline_benchmark_base.py
```diff
@@ -0,0 +1,198 @@
...
+    while self.num_workers <= self.max_num_replicas:
+
+      # Update the .yaml file replicas value
+      self._update_yaml(self.num_workers)
+      time.sleep(10)
+
+      # Delete and reapply the kubectl clusters
+      subprocess.call("kubectl delete -f beam_flink_cluster.yaml", shell=True)
+      subprocess.call("kubectl delete -f beam_job_server.yaml", shell=True)
+      time.sleep(180)
```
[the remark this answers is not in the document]
kshivvy: lolol I was conservative sleep time and I found it reliable up to 32 workers (haven't tested on 64). It'd be difficult to check in a loop since we'd have to parse kubectl output pretty

tfx/benchmarks/benchmark_base.py
kshivvy: I think to make them instance variables and methods I'd have to plumb the execution parameters through the tfma_benchmark_base.py, tfma_v2_benchmark_base.py, and tft_benchmark_base.py when they're first created since the beam pipeline is created via BenchmarkBase._create_beam_pipeline(). Let me know if that's the approach you want me to take.

tfx/benchmarks/beam_pipeline_benchmark_base.py
```diff
@@ -0,0 +1,198 @@
...
+import csv
+
+class BeamPipelineBenchmarkBase():
+  """Beam Pipeline benchmarks base class."""
+
+  def __init__(self, min_num_workers, max_num_workers, base_dir,
+               cloud_dataflow_temp_loc=""):
+
+    self.wall_times = {}
+    self.wall_times["tfma"] = {}
```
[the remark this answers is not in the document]
kshivvy: I'll stick with keys for now due to design considerations with scaling file size in the Big Shuffle benchmark, and I'll revisit the optional version once it's committed

tfx/benchmarks/beam_job_server.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: apps/v1beta1
```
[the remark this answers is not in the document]
kshivvy: I renamed both files as suggested.

tfx/benchmarks/beam_pipeline_benchmark_base.py
```diff
@@ -0,0 +1,198 @@
...
+    tft_benchmark_chicago_taxi = TFTBenchmarkChicagoTaxi(dataset=self.dataset)
+    self.wall_times["tft"]["TFTBenchmarkChicagoTaxi.benchmarkAnalyzeAndTransformDataset"] = tft_benchmark_chicago_taxi.benchmarkAnalyzeAndTransformDataset()
+
+  def _run_all_benchmarks(self):
+    self._run_tfma_benchmarks()
+    self._run_tfma_v2_benchmarks()
+    self._run_tft_benchmarks()
+
+  def _post_process(self):
+
```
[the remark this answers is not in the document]
kshivvy: Done.

tfx/benchmarks/benchmark_base.py
```diff
@@ -27,14 +28,73 @@
...
     # FLAGS may not be parsed if the benchmark is instantiated directly by a
     # test framework (e.g. PerfZero creates the class and calls the methods
     # directly)
     runner_flag = (
         FLAGS.beam_runner
         if FLAGS.is_parsed() else FLAGS["beam_runner"].default)
     return beam.Pipeline(runner=beam.runners.create_runner(runner_flag))
+
+  def _create_beam_pipeline(self):
+
```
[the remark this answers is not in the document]
kshivvy: Done.

tfx/benchmarks/benchmark_base.py
kshivvy: Changed to named constants.

tfx/benchmarks/benchmark_base.py
kshivvy: Changed to named constants.

tfx/benchmarks/benchmark_base.py
kshivvy: Understood, I was always curious about double vs single quotes haha. I switched everything to double quotes.

tfx/benchmarks/beam_pipeline_benchmark_chicago_taxi.py
kshivvy: I changed them to shell environment variables, maybe we can include a .sh script to set them up? I use one locally but not sure if its useful

tfx/benchmarks/beam_pipeline_benchmark_base.py
```diff
@@ -0,0 +1,184 @@
...
+import subprocess
+import time
+import yaml
+from absl import flags
+
+from datasets.chicago_taxi import dataset
+from tfma_benchmark_chicago_taxi import TFMABenchmarkChicagoTaxi
+from tfma_v2_benchmark_chicago_taxi import TFMAV2BenchmarkChicagoTaxi
+from tft_benchmark_chicago_taxi import TFTBenchmarkChicagoTaxi
+import benchmark_base
```
yifanmai: Use fully qualified imports instead rather than relative imports ([example](https://github.com/tensorflow/tfx/blob/de0e6c7316039ee5f3b2ab3e7814da2da223b8bd/tfx/benchmarks/tfma_benchmark_base.py#L29))

tfx/benchmarks/beam_pipeline_benchmark_chicago_taxi.py
```diff
@@ -0,0 +1,37 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""TFMA, TFMA v2, TFT Apache Beam pipeline benchmarks for Chicago Taxi dataset."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import sys
+sys.path.append("/home/kshivvy/tfx-benchmarks-PR/tfx/")
```
yifanmai: Hmm, is this needed to get things to work?

tfx/benchmarks/benchmark_base.py
```diff
@@ -27,14 +28,77 @@
...
 
+DEFAULT_MODE = "default"
+LOCAL_SCALED_EXECUTION_MODE = "local_scaled_execution"
+CLOUD_DATAFLOW_MODE = "cloud_dataflow"
+FLINK_ON_K8S_MODE = "flink_on_k8s"
+modes = [DEFAULT_MODE, LOCAL_SCALED_EXECUTION_MODE, CLOUD_DATAFLOW_MODE,
+         FLINK_ON_K8S_MODE]
+
+beam_pipeline_mode = DEFAULT_MODE
+num_workers_str = "1"
```
yifanmai, line 39: nit: make this an int, then use `"--num_workers=%d" % num_workers` to interpolate

tfx/benchmarks/big_shuffle_benchmark.py
```diff
@@ -0,0 +1,127 @@
...
+from tfx.benchmarks import benchmark_base
+
+class BigShuffleBenchmarkBase(benchmark_base.BenchmarkBase):
+  """Contains standalone Beam pipeline benchmarks, not dependent on outside
+     datasets and tensorflow dependencies."""
+
+  def __init__(self, file_size=1e6, input_file="input.txt",
+               output_file="output.txt"):
+    super(BigShuffleBenchmarkBase, self).__init__()
+    self.input_file = input_file
```
yifanmai, line 34: nit: for instance variables that are "private", prefix with `_` e.g. `self._input_file` as per [PEP 8](https://www.python.org/dev/peps/pep-0008/#id50).

tfx/benchmarks/big_shuffle_benchmark.py
```diff
@@ -0,0 +1,127 @@
...
+    bucket = storage_client.get_bucket(bucket_name)
+    blob = bucket.blob(destination_blob_name)
+
+    blob.upload_from_filename(self.tmp_input_file)
+    os.remove(self.tmp_input_file)
+
+  def _generate_data(self):
+
+    if self.read_write_to_cloud:
+      f = open(self.tmp_input_file, "a")
```
yifanmai, line 74: Prefer using [context manager style](https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files): `with open() as f`

This ensures that your file is closed after you're done.

Also, is there a use case for using 'a' instead of 'w' here? If this were 'w' then regenerate and generate wouldn't have to be treated differently.

tfx/benchmarks/big_shuffle_benchmark.py
```diff
@@ -0,0 +1,127 @@
...
+
+  def __init__(self, file_size=1e6, input_file="input.txt",
+               output_file="output.txt"):
+    super(BigShuffleBenchmarkBase, self).__init__()
+    self.input_file = input_file
+    self.output_file = output_file
+    self.file_size = file_size # 1e8 bytes = 100 MB
+
+    self.chars_per_line = 100
+    self.bytes_per_line = (self.chars_per_line * 2)
```
yifanmai, line 39: Is the `* 2` here correct? If you're using ASCII then one char = one byte.

(Also the newline takes up one byte, since these calculations are approximate it doesn't matter.)

tfx/benchmarks/big_shuffle_benchmark.py
```diff
@@ -0,0 +1,127 @@
...
+
+    self.tmp_input_file = "tmp_input.txt"
+    self.read_write_to_cloud = False
+    if self.input_file[0:5] == "gs://":
```
yifanmai, line 43: nit: `self.input_file.startswith("gs://")` ([doc](https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files))

tfx/benchmarks/big_shuffle_benchmark.py
```diff
@@ -0,0 +1,127 @@
...
+               output_file="output.txt"):
+    super(BigShuffleBenchmarkBase, self).__init__()
+    self.input_file = input_file
+    self.output_file = output_file
+    self.file_size = file_size # 1e8 bytes = 100 MB
+
+    self.chars_per_line = 100
+    self.bytes_per_line = (self.chars_per_line * 2)
+
+    self.tmp_input_file = "tmp_input.txt"
```
yifanmai, line 41: Use [`tempfile.NamedTemporaryFile`](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) instead, and also make this local to to `_generate_data`. You can pass it via parameter to `_upload_tmp_file_to_bucket`. In general, prefer passing values via parameter to passing values via instance variables.

tfx/benchmarks/benchmark_base.py
yifanmai: `TFMABenchmarkBase`, `TFMAV2BenchmarkBase` and `TFTBenchmarkBase` all subclass from `BenchmarkBase`, so you can make `set_beam_pipeline_mode` a method on `BenchmarkBase` and it would be accessible from those classes.

For example, in `_run_tfma_benchmarks()` you could then call `set_beam_pipeline_mode()` after constructing `TFMABenchmarkBase`. `_create_beam_pipeline()` can then read from this instance variable instead of from the global.

tfx/benchmarks/tfma_v2_benchmark_base.py
```diff
@@ -135,8 +136,8 @@ def benchmarkMiniPipelineBatched(self):
     pipeline = self._create_beam_pipeline()
     tfx_io = test_util.InMemoryTFExampleRecord(
         schema=benchmark_utils.read_schema(
-            self._dataset.tf_metadata_schema_path()),
-        raw_record_column_name=constants.ARROW_INPUT_COLUMN)
+            "../examples/chicago_taxi_pipeline/data/user_provided_schema/schema.pbtxt"),
+        raw_record_column_name='__raw_record__')
```
yifanmai, line 140: What do these changes do?

tfx/benchmarks/beam_pipeline_benchmark_base.py
```diff
@@ -0,0 +1,184 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import csv
+import os
+os.environ['TF_CPP_MIN_LOG_LEVEL'] = '3'
```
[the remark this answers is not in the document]
kshivvy: Removed the line, it was uneeded

tfx/benchmarks/beam_pipeline_benchmark_base.py
kshivvy: Done

tfx/benchmarks/beam_pipeline_benchmark_chicago_taxi.py
kshivvy: Nope it wasn't, I removed the line

tfx/benchmarks/benchmark_base.py
kshivvy: Done, made it an int.

tfx/benchmarks/big_shuffle_benchmark.py
kshivvy: Done, I went through the classes I made to check for private variables

tfx/benchmarks/big_shuffle_benchmark.py
kshivvy: You were right, 'w' worked fine. I switched to use `with open() as f`

tfx/benchmarks/big_shuffle_benchmark.py
kshivvy: You're right, no need for the `* 2`.

tfx/benchmarks/big_shuffle_benchmark.py
kshivvy: Done

tfx/benchmarks/big_shuffle_benchmark.py
kshivvy: I used a `tempfile.NamedTemporaryFile` instead. I'll try to pass values by parameter when possible moving forward.

tfx/benchmarks/tfma_v2_benchmark_base.py
kshivvy: For some reason, the dataset path stored in a cloud bucket for self._dataset.tf_metadata_schema_path() passed an path that was unreachable in the bucket, even though that exact path existed. I added the constant back in, I realized I had deleted an import that caused issues

kshivvy: For some reason, the dataset path stored in a cloud bucket for self._dataset.tf_metadata_schema_path() passed an path that was unreachable in the bucket, even though that exact path existed. I added the constant back in, I realized I had deleted an import that caused issues

tfx/benchmarks/beam_pipeline_benchmark_base.py
```diff
@@ -0,0 +1,198 @@
...
+      self._post_process()
+
+      # Write to csv
+      self._write_to_csv(csv_filename=
+                         "beam_pipeline_benchmark_results_flink_on_k8s.csv")
+
+      if self.num_workers < 4:
+        self.num_workers += 1
+      else:
+        self.num_workers *= 2
```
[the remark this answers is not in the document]
kshivvy: I made a helper function instead - let me know if you prefer the list parameter option and I can implement it that way

tfx/benchmarks/beam_pipeline_benchmark_base.py
```diff
@@ -0,0 +1,198 @@
...
+    # Add wall times to dataset
+    row = [self.num_workers]
+    for tf_module in self.wall_times:
+      for test, wall_time in self.wall_times[tf_module].items():
+        row.append(wall_time)
+
+    self.wall_times_list.append(row)
+    print(self.wall_times_list)
+
+  def _update_yaml(self, num_replicas):
```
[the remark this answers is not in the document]
kshivvy: I used the optional method and copied the yaml to a temporary file and modified it that way.

tfx/benchmarks/setup.py
```diff
@@ -0,0 +1,33 @@
...
+REQUIRED_PACKAGES = ["tfx==0.21.0",
+                     "tensorflow-transform==0.21.2",
+                     "tensorflow-model-analysis==0.21.4",
+                     "google-cloud-storage==1.29.0",
+                     "apache-beam==2.22.0"]
+
+# REQUIRED_PACKAGES = ["tfx==0.21.0",
+#                      "tensorflow-transform==0.21.2",
+#                      "tensorflow-model-analysis==0.21.4",
+#                      "apache-beam==2.18.0"]
```
yifanmai, line 26: Delete

tfx/benchmarks/beam_pipeline_benchmark_base.py
```diff
@@ -0,0 +1,253 @@
...
+TFMA_V2_BENCHMARK_MINI_PIPELINE_BATCHED = "TFMAV2BenchmarkChicagoTaxi.benchmarkMiniPipelineBatched"
+TFT_BENCHMARK_ANALYZE_AND_TRANSFORM_DATASET = "TFTBenchmarkChicagoTaxi.benchmarkAnalyzeAndTransformDataset"
+BIG_SHUFFLE_BENCHMARK = "BigShuffleBenchmarkBase.benchmarkBigShuffle with file size: "
+BIG_SHUFFLE_EMPTY_PIPELINE_BENCHMARK = "BigShuffleBenchmarkBase.benchmarkEmptyPipeline"
+
+
+class BeamPipelineBenchmarkBase():
+  """Beam Pipeline benchmarks base class."""
+
+  def __init__(self, min_num_workers, max_num_workers, base_dir,
```
yifanmai, line 50: I feel like many of the variables here could be parameters instead of instance variables:

- The runner-specific arguments could be moved to the runner-specific methods e.g. `cloud_dataflow_temp_loc` could be a parameter in `benchmarkCloudDataflow`; that way a user who wants to only run Flink benchmarks would not have to also configure Cloud Dataflow.
- It would be better for `num_workers` to be a local variable that is passed into only the methods that need them.

What do you think? This would require quite a bit of restructuring so I'm OK with leaving it out of this PR.

tfx/benchmarks/beam_pipeline_benchmark_base.py
```diff
@@ -0,0 +1,253 @@
...
+
+TFMA_BENCHMARK_MINI_PIPELINE = "TFMABenchmarkChicagoTaxi.benchmarkMiniPipeline"
+TFMA_V2_BENCHMARK_MINI_PIPELINE_UNBATCHED = "TFMAV2BenchmarkChicagoTaxi.benchmarkMiniPipelineUnbatched"
+TFMA_V2_BENCHMARK_MINI_PIPELINE_BATCHED = "TFMAV2BenchmarkChicagoTaxi.benchmarkMiniPipelineBatched"
+TFT_BENCHMARK_ANALYZE_AND_TRANSFORM_DATASET = "TFTBenchmarkChicagoTaxi.benchmarkAnalyzeAndTransformDataset"
+BIG_SHUFFLE_BENCHMARK = "BigShuffleBenchmarkBase.benchmarkBigShuffle with file size: "
+BIG_SHUFFLE_EMPTY_PIPELINE_BENCHMARK = "BigShuffleBenchmarkBase.benchmarkEmptyPipeline"
+
+
+class BeamPipelineBenchmarkBase():
```
yifanmai, line 47: nit: `class BeamPipelineBenchmarkBase(object)`

tfx/benchmarks/benchmark_base.py
```diff
@@ -19,22 +19,82 @@
...
+  """Base class for running Beam pipelines on various runners"""
 
-  def _create_beam_pipeline(self):
+  def __init__(self):
+    super(BenchmarkBase, self).__init__()
+    self.beam_pipeline_mode = mode_config.DEFAULT_MODE
+    self.num_workers = 1
+    self.cloud_dataflow_temp_loc = None
+
+  def _set_cloud_dataflow_flags(self):
```
yifanmai, line 41: Could you use [`PipelineOptions`](https://beam.apache.org/releases/pydoc/2.22.0/apache_beam.options.pipeline_options.html#apache_beam.options.pipeline_options.PipelineOptions) instead of flags? The flags get parsed into `PipelineOptions` by Beam anyway, so this gets rid of the need to do string construction and interpolation (which is more error prone).

tfx/benchmarks/benchmark_base.py
```diff
@@ -19,22 +19,82 @@
...
+  def _create_beam_pipeline_default(self):
     # FLAGS may not be parsed if the benchmark is instantiated directly by a
     # test framework (e.g. PerfZero creates the class and calls the methods
     # directly)
     runner_flag = (
         FLAGS.beam_runner
         if FLAGS.is_parsed() else FLAGS["beam_runner"].default)
     return beam.Pipeline(runner=beam.runners.create_runner(runner_flag))
+
+  def _create_beam_pipeline(self):
```
yifanmai, line 86: I'm concerned about the changes to the class API:

- This entangles details about Cloud Dataflow etc. into the Benchmark classes. This is not great because the benchmarks are about defining the _pipelines_ themselves rather than the _execution environment_. Keeping those two separate is one of the main philosophies of Beam.
- The runner configuration API is cumbersome: it allows users to set parameters that are unrelated to the runner e.g. I can call `set_cloud_dataflow_temp_loc` and then run the pipeline on Flink.

One way to maintain this separation is to add a `pipeline_options` parameter to `create_beam_pipeline` with a default value of `None`. Within `create_beam_pipeline`, pass those options into `beam.Pipeline`. Then provide a pure functions for each runner elsewhere that construct the appropriate `PipelineOptions`, with parameters appropriate for that runner.

What do you think? This would require quite a bit of restructuring so I'm OK with leaving it out of this PR.

tfx/benchmarks/beam_pipeline_benchmark_base.py
kshivvy: Done.

tfx/benchmarks/setup.py
kshivvy: Deleted.

tfx/benchmarks/benchmark_base.py
kshivvy: Done, I switched everything to use PipelineOptions instead of flags.

tfx/benchmarks/beam_pipeline_benchmark_base.py
kshivvy: I went through the variables that we discussed yesterday and made as many of them as possible local variables. For `cloud_dataflow_project` and `cloud_dataflow_temp_loc`, I set them as parameters to be passed into `benchmarkCloudDataflow` and removed them from the `BeamPipelineBenchmarkBase` constructor. However, I kept those two variables as instance variables  (i.e. with `self.`) because plumbing those as parameters through the methods that would require them made the code confusing and difficult to understand in my opinion. Let me know if you would like me to make those local as well.

yifanmai: This looks OK (since all of this classes' public methods require `big_shuffle_input_file` and `big_shuffle_output_file`.

tfx/benchmarks/beam_pipeline_benchmark_base.py
```diff
@@ -0,0 +1,269 @@
...
+    self.cloud_dataflow_project = None
+    self.cloud_dataflow_temp_loc = None
+
+  def _set_benchmark_class_parameters(self, benchmark_class, num_workers,
+                                      beam_pipeline_mode):
+    benchmark_class.set_num_workers(num_workers)
+    benchmark_class.set_beam_pipeline_mode(beam_pipeline_mode)
+
+    if beam_pipeline_mode == mode_config.CLOUD_DATAFLOW_MODE:
+      assert  self.cloud_dataflow_project
```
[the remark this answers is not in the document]
kshivvy: Done.

charlesccychen:
@kshivvy: Please merge the pull request https://github.com/kshivvy/tfx/pull/2 into your branch (which will update this pull request).

PR #2133, opened 2020-07-13
Supporting multi-artifact input channels to Trainer component

18jeffreyma:
Currently, Trainer assumes input channel only contains a single artifact, using `get_split_uri()` to parse out the artifact URI. To fix, added utility method for parsing all URIs from a list of artifacts, and updated `get_common_fn_args()` to pull glob patterns for all input artifacts for user.


tfx/components/trainer/fn_args_utils.py
```diff
@@ -59,14 +59,14 @@ def get_common_fn_args(input_dict: Dict[Text, List[types.Artifact]],
                        working_dir: Text = None) -> FnArgs:
   """Get common args of training and tuning."""
   train_files = [
-      io_utils.all_files_pattern(
-          artifact_utils.get_split_uri(input_dict[constants.EXAMPLES_KEY],
-                                       'train'))
+      io_utils.all_files_pattern(uri) for uri in
```
1025KB, line 62: could you check the api used in module file allows multiple file patterns?

(in case it only support single pattern or multiple specific files)

tfx/components/trainer/fn_args_utils.py
```diff
@@ -59,14 +59,14 @@ def get_common_fn_args(input_dict: Dict[Text, List[types.Artifact]],
...
+          artifact_utils.get_split_uris(input_dict[constants.EXAMPLES_KEY],
+                                        'train')
   ]
   eval_files = [
-      io_utils.all_files_pattern(
-          artifact_utils.get_split_uri(input_dict[constants.EXAMPLES_KEY],
-                                       'eval'))
+      io_utils.all_files_pattern(uri) for uri in
+          artifact_utils.get_split_uris(input_dict[constants.EXAMPLES_KEY],
+                                        'eval')
```
1025KB, line 69: can we have an executor_test case (new or update existing) to cover the code path

tfx/components/trainer/fn_args_utils.py
18jeffreyma: `trainer_module.py` uses `tf.data.experimental.make_batched_features_dataset()` to process a list of input file names. It accepts a list of patterns 

https://www.tensorflow.org/api_docs/python/tf/data/experimental/make_batched_features_dataset

tfx/components/trainer/fn_args_utils.py
18jeffreyma: added a test case, which specifically tests multiple artifact behavior

tfx/components/trainer/executor_test.py
```diff
@@ -123,6 +133,7 @@ def _do(self, test_executor):
         exec_properties=self._exec_properties)
 
   def testGenericExecutor(self):
+    self._input_dict[constants.EXAMPLES_KEY] = self._single_artifact
```
1025KB, line 136: you don't need to set this as it's default to single artifact right?

tfx/components/trainer/executor_test.py
```diff
@@ -47,20 +47,30 @@ def setUp(self):
...
+    e1 = standard_artifacts.Examples()
+    e1.uri = os.path.join(self._source_data_dir,
                                 'transform/transformed_examples')
-    examples.split_names = artifact_utils.encode_split_names(['train', 'eval'])
+    e1.split_names = artifact_utils.encode_split_names(['train', 'eval'])
+
+    e2 = standard_artifacts.Examples()
+    e2.uri = os.path.join(self._source_data_dir,
+                                'transform/transformed_examples')
+    e2.split_names = artifact_utils.encode_split_names(['train', 'eval'])
```
1025KB, line 58: wondering can we just do a copy instead of duplicated code

tfx/components/trainer/executor_test.py
18jeffreyma: one thing i wasn't sure about was the ordering of tests (whether they're run in order from top to bottom) and if setUp is called before each test. single artifact should be the default, but for test where I'm not checking for multiple artifact behavior, I added it to be explicit. should that be reverted?

tfx/components/trainer/executor_test.py
18jeffreyma: I wasn't sure if shallow copying is enough to test our multiple artifact (i.e. just using python assignment), which is why I decided on this way for now. is there some tfx utility for copying artifacts?

tfx/components/trainer/executor_test.py
18jeffreyma: resolved, was unclear on `setUp()` running before each test case (it does). will fix

tfx/components/trainer/executor_test.py
ruoyu90: we can use proto `CopyFrom()`. See [this](https://developers.google.com/protocol-buffers/docs/reference/python-generated) and search for 'CopyFrom'.

18jeffreyma: spoke with Ruoyu, decided to use `copy.deepcopy()` instead.

18jeffreyma:
Addressed comments, synced fork to tfx head.

18jeffreyma:
Currently, working on a test for the resolver in `iris_pipeline_resolver_e2e_test.py`. This will likely not merge with the other changes, but is pushed to this PR to test with our Trainer changes.

PR #2138, opened 2020-07-14
Adding Version support in ExampleGen component

18jeffreyma:
Currently, Span exists to represent groupings of data, but if a Span of data needs to be cleaned, there exists no way to track these changes beyond rerunning the entire pipeline on this changed data. 

This CL adds support for a new Example artifact property, Version to the ExampleGen component. The ExampleGen driver will attempt to find the latest Span, then find the latest Version for that Span, through parsing a split's pattern, and will provide the correct split patterns to the ExampleGen executor for processing and conversion to Example artifacts.


tfx/components/example_gen/driver_test.py
```diff
@@ -57,37 +57,40 @@ def testResolveExecProperties(self):
...
-    span1_split2 = os.path.join(self._input_base_path, 'span01', 'split2',
-                                'data')
-    io_utils.write_string_file(span1_split2, 'testing12')
-    span2_split1 = os.path.join(self._input_base_path, 'span02', 'split1',
-                                'data')
-    io_utils.write_string_file(span2_split1, 'testing21')
+    span1_v1_split1 = os.path.join(self._input_base_path, 'span01', 'version01',
+                                   'split1', 'data')
+    io_utils.write_string_file(span1_v1_split1, 'testing11')
+    span1_v1_split2 = os.path.join(self._input_base_path, 'span01', 'version01',
```
1025KB, line 73: let's have a v2, so we test the part of choosing latest version

tfx/components/example_gen/driver.py
```diff
@@ -55,13 +55,15 @@ def resolve_exec_properties(
     logging.debug('Processing input %s.', input_base)
 
     # Note that this function updates the input_config.splits.pattern.
-    fingerprint, select_span = utils.calculate_splits_fingerprint_and_span(
-        input_base, input_config.splits)
+    fp, span, version = utils.calculate_splits_fingerprint_span_and_version(
+                                input_base,
+                                input_config.splits)
```
1025KB, line 60: lint format

tfx/components/example_gen/utils.py
```diff
@@ -218,42 +249,49 @@ def _retrieve_latest_span(uri: Text,
...
   """Calculates the fingerprint of files in a URI matching split patterns.
 
-  If a pattern has the {SPAN} placeholder, attempts to find an identical value
-  across splits that results in all splits having the most recently updated
-  files.
+  If a pattern has the {SPAN} placeholder, attempts to find an identical
+  value across splits that results in all splits having the most recently
+  updated files.
+
+  If a pattern has the {VERSION} placeholder, attempts to, after finding span,
```
1025KB, line 292: duplicated `attempts to`?

tfx/components/example_gen/utils.py
```diff
@@ -262,16 +300,20 @@ def calculate_splits_fingerprint_and_span(
       Span number.
 
   Returns:
-    A Tuple of [fingerprint, select_span], where select_span is either
-    the value matched with the {SPAN} placeholder, or None if the placeholder
+    A Tuple of [fingerprint, select_span, select_version], where select_span
+    is either the value matched with the {SPAN} placeholder, or None if the
+    placeholder wasn't specified, and where select_version is either the
+    value matched with the {Version} placeholder, or None if the placeholder
```
1025KB, line 306: comment seems not matching the code

tfx/components/example_gen/utils.py
```diff
@@ -284,9 +326,26 @@ def calculate_splits_fingerprint_and_span(
...
+    # Given previous span, find most recent version
+    if VERSION_SPEC in split.pattern:
+      latest_version = _retrieve_latest_version_from_span(input_base_uri,
+                                                          split,
+                                                          select_span)
+      logging.info('latest version = %s', latest_version)
+      if select_version is None:
+        select_version = latest_version
+      if select_version != latest_version:
+        raise ValueError(
```
1025KB, line 339: we should have the same for span, cause the structure should be span/version/split, thus all split of the same artifact should have same span and version

tfx/components/example_gen/utils.py
```diff
@@ -218,42 +249,49 @@ def _retrieve_latest_span(uri: Text,
...
-      latest_span = result.group(1)
+  return _property_search(split, split_glob_pattern, split_regex_pattern,
+                          SPAN_PROPERTY_NAME, SPAN_SPEC)
 
-  if latest_span is None:
-    raise ValueError('Cannot not find matching for split %s based on %s' %
-                     (split.name, split.pattern))
-  return latest_span
 
+def _retrieve_latest_version_from_span(uri: Text,
```
1025KB, line 263: can we merge _retrieve_latest_version_from_span  & _retrieve_latest_span to reduce duplication?

tfx/components/example_gen/utils.py
18jeffreyma: i think Span logic should already be fine (unless it needs changes), since we check if our calculated `latest_span` is the same as the current `select_span`

https://github.com/18jeffreyma/tfx/blob/d7482f2502f4bc2472944e4cd1e40a7a7fc41b28/tfx/components/example_gen/utils.py#L280

18jeffreyma:
Addressed recent wave of comments, refactored span and version logic into smaller method, and fixed documentation issues.

tfx/components/example_gen/driver_test.py
```diff
@@ -100,18 +103,69 @@ def testResolveExecProperties(self):
...
+                        name='s1',
+                        pattern='span{SPAN}/version{VERSION}/split1/*'),
+                    example_gen_pb2.Input.Split(
+                        name='s2',
+                        pattern='span{SPAN}/version{VERSION}/split2/*')
+                ]),
+                preserving_proto_field_name=True),
+    }
+    
+    # Write new versions of files.
```
1025KB, line 130: I mean just have a v2 above, just run executor once is fine

(like above we have span1 and span2, span2 is picked up)

a single test can cover `pick up latest version of latest span`

tfx/components/example_gen/utils.py
```diff
@@ -209,51 +213,98 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
+  split.pattern = split.pattern.replace(SPAN_SPEC, latest_span)
+  split.pattern = split.pattern.replace(VERSION_SPEC, latest_version)
+
+  return latest_span, latest_version
+
+
+def calculate_splits_fp_span_and_version(
     input_base_uri: Text, splits: Iterable[example_gen_pb2.Input.Split]
-) -> Tuple[Text, Optional[Text]]:
+) -> Tuple[Text, Optional[Text], Optional[Text]]:
```
1025KB, line 298: it's not optional then

tfx/components/example_gen/utils_test.py
```diff
@@ -285,9 +311,47 @@ def testSpanMatches(self):
         example_gen_pb2.Input.Split(name='s1', pattern='span{SPAN}/split1/*'),
         example_gen_pb2.Input.Split(name='s2', pattern='span{SPAN}/split2/*')
     ]
-    _, span = utils.calculate_splits_fingerprint_and_span(
+    _, span, _ = utils.calculate_splits_fp_span_and_version(
         self._input_base_path, splits)
     self.assertEqual(span, '02')
+  
+  def testVersionMatches(self):
```
1025KB, line 318: consider just have just one test for testing calculate_splits_fp_span_and_version function (instead of testSpan and testVersion) to reduce some duplication

18jeffreyma:
Addressed recent wave of comments, thanks!

18jeffreyma:
There's some ImportError, which appeared when i tried syncing my changes to head. I'm not sure if tf.proto.orchestration was changed recently, but it seems to have broken my tests. Any idea on what could be causing it?

https://github.com/tensorflow/tfx/pull/2138/checks?check_run_id=871647666

18jeffreyma:
Synced fork to head. The ImportError on the tests on Github is due to an issue with the current Github workflow, but after rebuilding protos, all test cases pass.

tfx/components/example_gen/utils.py
```diff
@@ -209,51 +213,96 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
+                 split_regex_pattern)
+
+    latest_version = _property_search(split, split_glob_pattern,
+                                      split_regex_pattern,
+                                      VERSION_PROPERTY_NAME)
+
+  split.pattern = split.pattern.replace(SPAN_SPEC, latest_span)
+  split.pattern = split.pattern.replace(VERSION_SPEC, latest_version)
+
+  return latest_span, latest_version
```
hughmiao, line 291: can we use up to two capture groups? 
if span is given, but not version, we use one for span, return span number, and nil for version
if span/version are given, we use two,  return a pair of numbers, error out if number is invalid
if span is not given, version is given, we error out. 

by doing so, we avoid glob twice, and we can combine the _property_search in a cleaner way to update the tuple together.

tfx/components/example_gen/utils.py
```diff
@@ -209,51 +213,96 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
+
+
+def _retrieve_latest_span_version(uri: Text,
+                                  split: example_gen_pb2.Input.Split
+                                  ) -> Tuple[Text, Text]:
+  """Retrieves the most recent span and version for a given split pattern"""
+  split_pattern = os.path.join(uri, split.pattern)
+
+  latest_span = '0'
+  latest_version = '0'
```
hughmiao, line 251: version can be optional

tfx/components/example_gen/utils.py
```diff
@@ -262,31 +311,38 @@ def calculate_splits_fingerprint_and_span(
       Span number.
 
   Returns:
-    A Tuple of [fingerprint, select_span], where select_span is either
-    the value matched with the {SPAN} placeholder, or None if the placeholder
+    A Tuple of [fingerprint, select_span, select_version], where select_span
+    is either the value matched with the {SPAN} placeholder, or '0' if the
+    placeholder wasn't specified, and where select_version is either the
+    value matched with the {VERSION} placeholder, or '0' if the placeholder
```
hughmiao, line 317: shall we distinguish a span = 0, vs. a span is None as before, similar to version.
we let the callsite of this function decide how to interpret None.

tfx/components/example_gen/utils.py
```diff
@@ -209,51 +213,96 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
+    logging.info('Span regex pattern for split %s: %s', split.name,
+                 split_regex_pattern)
 
+    latest_span = _property_search(split, split_glob_pattern,
+                                   split_regex_pattern, SPAN_PROPERTY_NAME)
 
-def calculate_splits_fingerprint_and_span(
+  if VERSION_SPEC in split.pattern:
+    if split.pattern.count(VERSION_SPEC) != 1:
+      raise ValueError('Only one {VERSION} is allowed in %s' % split_pattern)
```
hughmiao, line 272: nit, move the validation upfront

tfx/components/example_gen/utils.py
```diff
@@ -209,51 +213,96 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
     raise ValueError('Cannot not find matching for split %s based on %s' %
                      (split.name, split.pattern))
-  return latest_span
+  return latest_property
+
+
+def _retrieve_latest_span_version(uri: Text,
+                                  split: example_gen_pb2.Input.Split
+                                  ) -> Tuple[Text, Text]:
+  """Retrieves the most recent span and version for a given split pattern"""
```
hughmiao, line 247: nit, update the docstring given the cases is more complex than before.

tfx/components/example_gen/driver.py
```diff
@@ -55,13 +55,15 @@ def resolve_exec_properties(
     logging.debug('Processing input %s.', input_base)
 
     # Note that this function updates the input_config.splits.pattern.
-    fingerprint, select_span = utils.calculate_splits_fingerprint_and_span(
-        input_base, input_config.splits)
+    fp, span, version = utils.calculate_splits_fp_span_and_version(
```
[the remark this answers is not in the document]
18jeffreyma: reverted, thanks!

tfx/components/example_gen/utils.py
18jeffreyma: the main rationale for my decision to do glob twice is I was thinking about edge cases like if we had a pattern `{VERSION}/{SPAN}/...`. is assuming an ordering of span then version in the split config something we can do?

tfx/components/example_gen/utils.py
18jeffreyma: agreed, i've updated the `calculate_splits_fp_span_and_version` to interprete what a None property means.

tfx/components/example_gen/utils.py
18jeffreyma: moved the path logic into a function, which moves the validation up

tfx/components/example_gen/utils.py
18jeffreyma: added a doc string to `_retrieve_latest_span_version()`.

tfx/components/example_gen/utils.py
18jeffreyma: after working at this for a bit, i realize that we can use named groups to specifically parse our span and version, regardless of the order. fixed it so that only one glob pass is needed.

18jeffreyma:
Resolved all above comments, thanks! Some recent changes to note:

- Property search has been reworked to now be able to take in a list of properties and their spec (in decreasing order of importance) and retreive a list of the latest properties for each. I believe this implementation will be helpful if another similar property to Version/Span is added in the future.

- `_retrieve_latest_span_version()` now throws an error if Span is not present but Version is.

- `calculate_splits_fingerprint_span_and_version()` now takes care of interpreting what a None value for a property means, and assigns a span/version of '0' if `_retrieve_latest_span_version()` returns None.

tfx/components/example_gen/driver_test.py
```diff
@@ -57,61 +57,82 @@ def testResolveExecProperties(self):
             json_format.MessageToJson(
                 example_gen_pb2.Input(splits=[
                     example_gen_pb2.Input.Split(
-                        name='s1', pattern='span{SPAN}/split1/*'),
+                        name='s1',
+                        pattern='span{SPAN}/version{VERSION}/split1/*'),
```
hughmiao, line 61: let's write the behaviors test for

1. have {span}, not {version}
2. have both {span} and {version}
3. have {version}, but not {span} 
4. have none of them

tfx/components/example_gen/utils.py
```diff
@@ -209,84 +213,180 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
+
+  if SPAN_SPEC in split.pattern:
+    if split.pattern.count(SPAN_SPEC) != 1:
+      raise ValueError('Only one {SPAN} is allowed in %s' % split_pattern)
+
+    if VERSION_SPEC in split.pattern:
+      if split.pattern.count(VERSION_SPEC) != 1:
+        raise ValueError('Only one {VERSION} is allowed in %s' % split_pattern)
+
+      latest_span, latest_version = _property_search(uri, split,
```
hughmiao, line 318: nit, rename _property_search to `_glob_and_match_span_version`, which may read better

tfx/components/example_gen/utils.py
```diff
@@ -209,84 +213,180 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
   return regex_pattern
 
-
-def _retrieve_latest_span(uri: Text,
-                          split: example_gen_pb2.Input.Split) -> Text:
-  """Retrieves the most recently updated span matching a given split pattern."""
+def _property_search(uri: Text,
+                     split: example_gen_pb2.Input.Split,
+                     property_names: List[Text],
+                     spec_list: List[Text]) -> List[Text]:
```
hughmiao, line 219: given span/version are fundamental concepts in tfx data organization, it's unlikely we will extend to other matching groups at the moment. does it make sense to simply passing a boolean bit, e.g., is_match_version, instead of two index aligned lists? so the rest of the code can simply use SPAN_SPEC, VERSION_SPEC as constants instead of passing the two list. My concern is this looks like a premature design to handle the general case. wdyt?

tfx/components/example_gen/utils.py
18jeffreyma: fixed

tfx/components/example_gen/utils.py
18jeffreyma: I can rewrite it to have that structure instead. my only preference for this way was (1) the previous potential extensibility (if more specs are added later, which is unlikely) and (2) less casework required in the `_property_search` method. 

It seems like we already do casework in whether span is present and whether version is present (or both), so adding this boolean seems like it produces additional casework that wasn't needed before.

However, because we're renaming this to `_glob_and_match_span_version` (more specialized function), I think this design is probably not really all that necessary. will try an implementation with a boolean.

tfx/components/example_gen/driver_test.py
18jeffreyma: you left this comment on the driver_tests, is that where these test scenarios should go, or would they go in the util_test file?

18jeffreyma:
addressed most recent round of comments, thanks! notable changes:

- `_property_search()` has been renamed to `_glob_and_match_span_version()` and now uses a boolean flag to determine if version will be searched
- `utils_test.py` coverage increased to cover four scenarios: both span and version present; span present, but no version; version present but no span; and neither span nor version present

tfx/components/example_gen/utils.py
18jeffreyma: latest commits should have this boolean implementation

tfx/components/example_gen/driver_test.py
18jeffreyma: as of now, I put these tests under `utils_test.py`, since the behavior we're testing is specifically in `utils.py`

18jeffreyma:
Some additional changes:

- Some unit tests for testing the Span/Version matching behavior were considering multiple splits (when only one split was created in the temp directory). Unnecessary splits were removed from a couple unit test.

- Minor checks and bug fixes

tfx/components/example_gen/utils.py
```diff
@@ -209,84 +213,186 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
+
+    if VERSION_SPEC in split.pattern:
+      if split.pattern.count(VERSION_SPEC) != 1:
+        raise ValueError('Only one %s is allowed in %s' % (VERSION_SPEC,
+                                                           split.pattern))
+      latest_span, latest_version = _glob_and_match_span_version(uri, split,
+                                                                 True)
+
+    else:
+      latest_span, _ = _glob_and_match_span_version(uri, split, False)
```
hughmiao, line 330: nit, 

is_match_verison = VERSION_SPEC in split.pattern
if is_match_version and split.patttern.cout ..
     raise ValueError..

latest_span, latest_version = _glob...(uri, split, is_match_version)

tfx/components/example_gen/utils.py
```diff
@@ -209,84 +213,186 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
+      except ValueError:
+        raise ValueError('Cannot find %s number from %s based on %s' %
+                         (VERSION_PROPERTY_NAME, file_path,
+                          split_regex_pattern))
+
     if latest_span is None or span >= int(latest_span):
-      # Uses str instead of int because of zero padding digits.
-      latest_span = result.group(1)
+      latest_span = select_span
+      latest_version = select_version
```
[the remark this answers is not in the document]
18jeffreyma: reverted that comment removal, and put it closer to the assignment

18jeffreyma:
addressed most recent feedback, thanks!

tfx/components/example_gen/utils.py
```diff
@@ -209,84 +213,183 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
+
+  latest_span = None
+  latest_version = None
+
+  if SPAN_SPEC in split.pattern:
+    if split.pattern.count(SPAN_SPEC) != 1:
+      raise ValueError('Only one %s is allowed in %s' % (SPAN_SPEC,
+                                                         split.pattern))
+
+    is_match_version = VERSION_SPEC in split.pattern
```
ruoyu90, line 321: Will there be any problem if we always optimistically try to get the version info? We don't need the `is_match_version` then.

18jeffreyma: I think that it would just move this `is_match_version` boolean inside the `_glob_and_match_span_version()`. Since we already do all the processing to check if the `{VERSION}` spec is present in  `_retrieve_latest_span_version()`, I think that would repeat code if we were to optimistically try to get version info in the  `_glob_and_match_span_version()`.

On a structure note, I think having `_retrieve_latest_span_version()` as verifying the specs in the pattern and deciding what to match in the glob (like it does right now through the boolean) makes more sense than moving the spec verification inside  `_glob_and_match_span_version()`. @hughmiao any opinion?

tfx/components/example_gen/utils.py
```diff
@@ -209,84 +213,183 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
+                                                               is_match_version)
+
+  elif VERSION_SPEC in split.pattern:
+    raise ValueError('Version spec provided, but Span spec is not present')
+
+  # Replace split.pattern so executor can find files after driver runs.
+  if latest_span:
+    split.pattern = split.pattern.replace(SPAN_SPEC, latest_span)
+  if latest_version:
+    split.pattern = split.pattern.replace(VERSION_SPEC, latest_version)
```
1025KB, line 336: I feel we can move this to calculate_splits_fingerprint_span_and_version, and keep this function as a read only function?

tfx/components/example_gen/utils.py
1025KB: I feel we can merge this function and _glob_and_match_span_version

if we move line 332~336 to calculate_splits_fingerprint_span_and_version (check the comments below)

this function is basically just check existance and count, and it's pretty much align with the code structure in _glob_and_match_span_version line 223~234, consider merge them?

tfx/components/example_gen/utils.py
```diff
@@ -209,84 +213,183 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
+        '(?P<{}>.*)'.format(VERSION_PROPERTY_NAME))
+
   logging.info('Glob pattern for split %s: %s', split.name, split_glob_pattern)
-  split_regex_pattern = _glob_to_regex(split_pattern).replace(SPAN_SPEC, '(.*)')
   logging.info('Regex pattern for split %s: %s', split.name,
                split_regex_pattern)
-  if re.compile(split_regex_pattern).groups != 1:
-    raise ValueError('Regex should have only one group')
+
+  num_groups = re.compile(split_regex_pattern).groups
```
1025KB, line 239: do we assume span is always the first group and version is the second? if so we should assert that

tfx/components/example_gen/utils.py
```diff
@@ -209,84 +213,183 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
   files = tf.io.gfile.glob(split_glob_pattern)
   latest_span = None
+  latest_version = None
+
   for file_path in files:
     result = re.search(split_regex_pattern, file_path)
     if result is None:
       raise ValueError('Glob pattern does not match regex pattern')
+
+    select_span = result.group(SPAN_PROPERTY_NAME)
```
1025KB, line 254: consider call it 

latest_span/result_span, span_str and span_int as current iteration's span

latest_span, select_span, span feels confusing

tfx/components/example_gen/utils.py
```diff
@@ -209,84 +213,183 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
+                         (VERSION_PROPERTY_NAME, file_path,
+                          split_regex_pattern))
+
     if latest_span is None or span >= int(latest_span):
       # Uses str instead of int because of zero padding digits.
-      latest_span = result.group(1)
+      latest_span = select_span
+      latest_version = select_version
+    if latest_version is None or version >= int(latest_version):
+      latest_version = select_version
```
1025KB, line 276: how about this? I feel the logic is more straight forward:

if latest_span is None or span > int(latest_span)
      latest_span = select_span
      latest_version = select_version
elif span == int(latest_span) and version > int(latest_version):
      latest_version = select_version

and line275~276 seems will update other span's version to latest span, why test doesn't catch this?

tfx/components/example_gen/utils.py
18jeffreyma: moved this to `calculate_splits_fp_span_and_ver4sion()`

tfx/components/example_gen/utils.py
18jeffreyma: I spoke with Hui about this, and was told that there aren't any guarantees that Span is first group and Version is second (i.e. a pattern like `ver{VERSION}/span{SPAN}/*`. We don't assume this in our implementation since we're using named groups to retrieve our Span and Version implementation rather than relying on order.

tfx/components/example_gen/utils.py
18jeffreyma: yep, it should be elif, I will add a specific test case for this scenario.

tfx/components/example_gen/utils.py
18jeffreyma: renamed

tfx/components/example_gen/utils.py
```diff
@@ -209,84 +213,183 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
-  files.
+  If a pattern has the {SPAN} placeholder and, optionally, the {VERSION}
+  placeholder, attempts to find aligned values that results in all splits
+  having the most recent span and most recent version for that span.
 
   Args:
-    input_base_uri: The base path from which files will be searched
+    input_base_uri: The base path from which files will be searched.
     splits: An iterable collection of example_gen_pb2.Input.Split objects. Note
       that this function will update the {SPAN} in this split config to actual
```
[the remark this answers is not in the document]
18jeffreyma: updated docstring

tfx/components/example_gen/utils.py
18jeffreyma: latest draft of PR has them merged now

tfx/components/example_gen/utils.py
1025KB: my bad, I noticed that but forgot to remove my comment

tfx/components/example_gen/utils.py
```diff
@@ -210,83 +214,171 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
-  if latest_span is None:
-    raise ValueError('Cannot not find matching for split %s based on %s' %
-                     (split.name, split.pattern))
-  return latest_span
-
-
-def calculate_splits_fingerprint_and_span(
+  latest_version = None
+
+  if SPAN_SPEC in split.pattern:
```
1025KB, line 252: consider this, thus the `rest of the code` doesn't have indentation, and you don't need to have the else that too far from if to improve readability , wdyt?
```
if SPAN_SPEC not in split.pattern:
   if VERSION_SPEC in split.pattern
      raise ValueError
   return xxx

rest of the code
```

tfx/components/example_gen/utils.py
```diff
@@ -210,83 +214,171 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
+    logging.info('Glob pattern for split %s: %s', split.name,
+                 split_glob_pattern)
+    logging.info('Regex pattern for split %s: %s', split.name,
+                 split_regex_pattern)
+
+    num_groups = re.compile(split_regex_pattern).groups
+    if is_match_version and num_groups != 2:
+      raise ValueError('Span and version regex should have two groups')
+    elif not is_match_version and num_groups != 1:
+      raise ValueError('Span only regex should have one group')
```
1025KB, line 279: this seems duplicate with about split.pattern.count check?

(if you want to double check, you can do `assert` cause it seems a bug at our side if this triggers, not sth users need to fix)

18jeffreyma: yep its remainder from the old glob function, checked it with assert, ill fix that

tfx/components/example_gen/utils.py
18jeffreyma: agreed, the elif is very hard to read. fixed

tfx/components/example_gen/utils.py
```diff
@@ -210,83 +214,171 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
+                                                                split)
+
+    # Replace split.pattern so executor can find files after driver runs.
+    if latest_span:
+      split.pattern = split.pattern.replace(SPAN_SPEC, latest_span)
+    if latest_version:
+      split.pattern = split.pattern.replace(VERSION_SPEC, latest_version)
+
+    latest_span = latest_span or '0'
+    latest_version = latest_version or '0'
```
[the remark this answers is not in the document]
18jeffreyma: `set_string_custom_property()` expects a byte or unicode string. I've updated it to store the empty string instead of None.

18jeffreyma:
Addressed recent round of comments thanks! Notable changes:
- Fixed a bug where a newer span would potentially have a new version number updated from an older span.
- Updated proto documentation
- When Version spec is not present, `calculate_fingerprint_span_and_version()` now returns None for Version. Additionally, if Version is none, the Version custom property for the example artifact is set to the empty string `''`.

tfx/components/example_gen/utils.py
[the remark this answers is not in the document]
18jeffreyma: update: if version is None, behavior is now to not set the custom property. in any case, the value of that string custom property defaults to `''`.

18jeffreyma:
Updated PR to not set version property if None, and synced to TFX head for merging.

PR #2187, opened 2020-07-20
Support Custom Split Config for TFMA Component

tinally:
# Support custom split config for Evaluator component by adding examples_path_splits param.


tfx/components/evaluator/executor.py
```diff
@@ -176,9 +180,21 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
           eval_saved_model_path=model_path,
           add_metrics_callbacks=add_metrics_callbacks))
 
-    file_pattern = io_utils.all_files_pattern(
-        artifact_utils.get_split_uri(input_dict[constants.EXAMPLES_KEY], 'eval')
-    )
+    # Load and deserialize examples path splits from execution properties.
+    examples_path_splits = json_utils.loads(
+        exec_properties.get(constants.EXAMPLES_PATH_SPLITS_KEY)) or []
```
1025KB, line 185: `json_utils.loads(exec_properties.get(xxx, 'null')) or []` is correct:

exec_properties.get('xxx', 'null') will return 'null' when exec_properties['xxx'] = None
json_utils.loads('null') will return None, then None or [] will be []

tfx/components/evaluator/executor.py
```diff
@@ -213,9 +234,11 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
              slice_spec=slice_spec,
              tensor_adapter_config=tensor_adapter_config))
       else:
-        data = pipeline | 'ReadFromTFRecord' >> beam.io.ReadFromTFRecord(
-            file_pattern=file_pattern)
-        (data
+        for file_pattern in file_patterns:
```
1025KB, line 237: this else block can be merged with about else in line 223?

tfx/components/evaluator/executor_test.py
```diff
@@ -75,6 +76,10 @@ def testEvalution(self, exec_properties):
         constants.BLESSING_KEY: [blessing_output],
     }
 
+    # List needs to be serialized before being passed into Do function.
+    exec_properties[constants.EXAMPLES_PATH_SPLITS_KEY] = json_utils.dumps(
+        ['eval'])
```
1025KB, line 81: could you verify if multiple splits works (no need to update this PR though, manually try ['train', 'eval'] to see if it works) ?

tfx/components/evaluator/component.py
```diff
@@ -83,6 +84,7 @@ def __init__(
                                            Dict[Text, Any]]] = None,
       fairness_indicator_thresholds: Optional[List[Union[
           float, data_types.RuntimeParameter]]] = None,
+      examples_path_splits: Optional[List[Text]] = None,
```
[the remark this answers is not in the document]
tinally: should we align with the internal though? i think the internal code you pasted uses this name. though i agree this name is long and vague.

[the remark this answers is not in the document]
1025KB: could you check with Mike see if we can change the name or it's preferred to use the same name?

I think we do update some bad names during oss

tfx/components/evaluator/executor_test.py
```diff
@@ -132,7 +137,7 @@ def testDoLegacySingleEvalSavedModelWFairness(self, exec_properties):
       # post-export metric is registered.  This may raise an ImportError if the
       # currently-installed version of TFMA does not support fairness
       # indicators.
-      import tensorflow_model_analysis.addons.fairness.post_export_metrics.fairness_indicators  # pylint: disable=g-import-not-at-top, unused-variable
+      import tensorflow_model_analysis.addons.fairness.post_export_metrics.fairness_indicators  # pylint: disable=g-import-not-at-top, unused-variable, import-outside-toplevel, unused-import
```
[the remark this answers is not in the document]
tinally: in the comments, it mentioned when the version of TFMA supports fairness indicators, this import will be needed?

tfx/components/evaluator/component.py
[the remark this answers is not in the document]
tinally: got it, i will confirm with Mike about this

tfx/components/evaluator/executor_test.py
tinally: yes, it works with multiple splits now, i tried locally

tfx/components/evaluator/executor.py
tinally: i understand the difference now, i will correct the other components

tfx/components/evaluator/executor_test.py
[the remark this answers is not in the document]
1025KB: ah I think you are correct, 

there are 4 disables: g-import-not-at-top, unused-variable, import-outside-toplevel, unused-import
I think we only need two of them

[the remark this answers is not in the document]
tinally: i added the last two because the pylint failed when i didn't have all 4, if that's ok, i will remove the last 2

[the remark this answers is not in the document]
1025KB: g-import-not-at-top ～= import-outside-toplevel
unused-variable ~= unused-import

so I guess we only need 2, could you try remove the g-import-not-at-top, unused-variable see if it works for your PR

tfx/components/evaluator/executor.py
```diff
@@ -213,9 +235,7 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
              slice_spec=slice_spec,
              tensor_adapter_config=tensor_adapter_config))
       else:
-        data = pipeline | 'ReadFromTFRecord' >> beam.io.ReadFromTFRecord(
-            file_pattern=file_pattern)
-        (data
+        (examples_list | 'FlattenExamples' >> beam.Flatten()
```
1025KB, line 238: this seems incorrect, examples_list is empty in this else branch (where is beam.io.ReadFromTFRecord?)

tfx/components/evaluator/executor.py
```diff
@@ -213,9 +235,7 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
              slice_spec=slice_spec,
              tensor_adapter_config=tensor_adapter_config))
       else:
```
1025KB, line 237: this 215 else branch is the same as 204 else branch:
so no need to have 

```
if _USE_TFXIO:
  if tfma.is_batched_input(eval_shared_model, eval_config):
  else:
else:
```

it can be just

```
if _USE_TFXIO and tfma.is_batched_input(eval_shared_model, eval_config):
else:
```

tfx/components/evaluator/executor.py
tinally: There's no else branch here anymore. beam.io.ReadFromTFRecord is in line 199.

tfx/components/evaluator/executor.py
tinally: I already removed the first else, so it's like this now. I thought it will still be 2 lines so I didn't use `and` to combine _USE_TFXIO and is_batched_input.

```
if _USE_TFXIO:
  if tfma.is_batched_input(eval_shared_model, eval_config):
else:
```

1025KB: the tfma.ExtractEvaluateAndWriteResults call can be dedup too

```
examples_list = []
tensor_adapter_config = None
if xxx:
  tensor_adapter_config = xxx
  examples_list=xxx
else:
  examples_list=xxx

tfma.ExtractEvaluateAndWriteResults(...)
```

tfx/components/evaluator/executor_test.py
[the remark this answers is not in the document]
tinally: i removed the first two (g-import-not-at-top and unused-variable) and it works.

tfx/components/evaluator/executor.py
tinally: ah i see, yes, this is even better, i updated it

tfx/components/evaluator/component.py
```diff
@@ -142,6 +147,11 @@ def __init__(
       absl.logging.warning('feature_slicing_spec is deprecated, please use '
                            'eval_config instead.')
 
+    if not example_splits:
+      example_splits = ['eval']
+      absl.logging.info("Computing metrics on the 'eval' split when "
```
mdreves, line 152: because

tfx/components/evaluator/component.py
```diff
@@ -142,6 +147,11 @@ def __init__(
...
+                        "example_splits is not set.")
```
mdreves, line 153: The 'example_splits' parameter is not set, using 'eval' split.

tfx/components/evaluator/executor.py
```diff
@@ -176,52 +180,60 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+
     absl.logging.info('Evaluating model.')
     with self._make_beam_pipeline() as pipeline:
+      examples_list = []
+      tensor_adapter_config = None
       # pylint: disable=expression-not-assigned
-      if _USE_TFXIO:
-        tensor_adapter_config = None
-        if tfma.is_batched_input(eval_shared_model, eval_config):
+      if _USE_TFXIO and tfma.is_batched_input(eval_shared_model, eval_config):
```
mdreves, line 202: We seem to be missing the case where _USE_TFXIO is True, but tfma.is_batched_input(eval_shared_model, eval_config) is False.

tfx/components/evaluator/executor_test.py
```diff
@@ -28,6 +28,7 @@
 from tfx.proto import evaluator_pb2
 from tfx.types import artifact_utils
 from tfx.types import standard_artifacts
+from tfx.utils import json_utils
```
mdreves, line 33: Do we have tests with multiple splits?

tfx/components/evaluator/executor.py
tinally: that case is combined in the else branch, because when _USE_TFXIO is True but is_batched_input is False, tensor_adapter_config is still None, which is equivalent to the last else branch. so we decided to merge them.

tfx/components/evaluator/executor.py
```diff
@@ -176,52 +180,60 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
-             eval_config=eval_config,
-             output_path=output_uri,
-             slice_spec=slice_spec))
+        for split in example_splits:
+          file_pattern = io_utils.all_files_pattern(
+              artifact_utils.get_split_uri(input_dict[constants.EXAMPLES_KEY],
+                                           split))
+          data = (pipeline
+                  | 'ReadFromTFRecord[%s]' % split >>
+                  beam.io.ReadFromTFRecord(file_pattern=file_pattern))
```
1025KB, line 226: check this API [ReadAllFromTFRecord](https://github.com/apache/beam/blob/b91560cc354da471e3de502aad78dd059997a3d0/sdks/python/apache_beam/io/tfrecordio.py#L206), we might be able to use this

tfx/components/evaluator/executor.py
```diff
@@ -176,52 +180,60 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
-             output_path=output_uri,
-             slice_spec=slice_spec,
-             tensor_adapter_config=tensor_adapter_config))
+          data = (pipeline
+                  | 'ReadFromTFRecordToArrow[%s]' % split >>
+                  tfxio.BeamSource())
+          examples_list.append(data)
+        if schema is not None:
+          tensor_adapter_config = tensor_adapter.TensorAdapterConfig(
+              arrow_schema=tfxio.ArrowSchema(),
```
1025KB, line 217: just noticed this depends on the field in the for loop, then it might not be correct, we probably need to enable tf_example_record.TFExampleRecord to take a list of file_pattern first, let's check with Zhuo about this

tinally: yes, i recognized arrow_schema depends on tfxio and here it's always using the tfxio with the last file pattern in the list. 

if tf_example_record.TFExampleRecord can take in a list of file_pattern, then this tfxio will be correct to generate the arrow_schema.

tfx/components/evaluator/executor_test.py
tinally: i updated the first test to include both 'train' and 'eval' for example_splits. is this what you were looking for?

tfx/components/evaluator/component.py
mdreves: Is this changed? Another option is to replace 'when' with 'because'

tinally: yes, this is changed, try refresh the code or page, you should be able to see the change

tfx/components/evaluator/executor.py
1025KB: let's try this ReadAllFromTFRecord (check [here](https://github.com/apache/beam/search?q=ReadAllFromTFRecord&unscoped_q=ReadAllFromTFRecord) for the usage) for the else branch

line 234 need to be added to the if branch when we use  ReadAllFromTFRecord

tfx/components/evaluator/executor.py
```diff
@@ -36,6 +36,7 @@
 from tfx.types import artifact_utils
 from tfx.utils import io_utils
 from tfx.utils import path_utils
+from tfx.utils import json_utils
```
1025KB, line 39: move json_utils before path_utils

PR #2190, opened 2020-07-21
Enabling analyzer cache optimization in Transform component

18jeffreyma:
Analyzer caching is disabled at the moment since no artifact exists in the component spec to give a working directory to write cache information. After changes, a new `TransformCache` artifact type is created, and the Transform component optionally takes in a input channel of type `TransformCache` (if cache exists) and, in addition to current outputs, also outputs a `TransformCache` output artifact, where caching information is stored.

Since optional output artifacts are not yet possible and since caching information is calculated regardless of whether a cache output artifact to write to is provided, current behavior is to use a input cache artifact if provided, and to always provide a cache output artifact.


tfx/components/transform/executor_test.py
```diff
@@ -206,49 +207,33 @@ def testCounters(self):
   def testDoWithCache(self):
 
     # First run that creates cache.
-    output_cache_artifact = _OutputCache()
```
1025KB, line 209: so the cache is already supported in do(), just it is not exposed to component level?

tfx/components/transform/component.py
```diff
@@ -103,6 +105,10 @@ def preprocessing_fn(inputs: Dict[Text, Any]) -> Dict[Text, Any]:
       transformed_examples: Optional output 'ExamplesPath' channel for
         materialized transformed examples, which includes both 'train' and
         'eval' splits.
+      cache_input_path: Optional input 'TransformCache' channel containing
+        cached information from previous Transform runs.
```
1025KB, line 109: could you check the cache folder structure? I feel it might need some change

e.g., 

run 1: tft takes span1, span2, span3, and generate cache for 123
run 2: tft takes span2, span3, span4, and generate cache for 234

for run2, it take the cache from resovler, which contains 1,2,3 (only 2 and 3 is useful, I wonder if it is already that smart to record the span information in cache outputs)

and the worse case, resolver of cache is not align with resolver of examples, e.g., example resolver returns 3 4 5, cache resolver returns 6 7 8, will it crash?

1025KB: before cache information is store with output artifact, so it might be 1:1 mapping.

the cache channel might also has multiple artifacts

tfx/components/transform/executor_test.py
18jeffreyma: based on my current understanding yep. `cache_input_path` key was never provided in input_dict nor `cache_output_path` in output_dict since neither existed in the Transform component spec (thus why caching was not enabled yet)

tfx/components/transform/component.py
18jeffreyma: Resolved over chat conversation (i think), as Cache artifact stores whether a pattern has already been processed, not how a span has been prcession. Will resolve this once we hear back from Zohar.

18jeffreyma: Zohar confirmed that Cache artifact is a DatasetKey (file pattern) to cache info structure (i.e. one output artifact). Resolved.

18jeffreyma:
Currently, waiting to see Charles' soln for the optional artifact output, will revisit this when that is resolved.

18jeffreyma:
From syncing with Jiayi, decided that this PR will just enable it (always write cache). When optional output artifacts are enabled, another PR to incorporate that change will be submitted.

tfx/components/transform/executor.py
```diff
@@ -283,6 +283,8 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
           Tensorflow graph suitable for both training and serving;
         - transformed_examples: Materialized transformed examples, which
           includes both 'train' and 'eval' splits.
+        - cache_output_path: Cache output of 'tf.Transform', where cached
```
1025KB, line 286: cache_input_path?

tfx/components/transform/component.py
```diff
@@ -125,6 +131,9 @@ def preprocessing_fn(inputs: Dict[Text, Any]) -> Dict[Text, Any]:
     transform_graph = transform_graph or types.Channel(
         type=standard_artifacts.TransformGraph,
         artifacts=[standard_artifacts.TransformGraph()])
+    cache_output_path = cache_output_path or types.Channel(
+        type=standard_artifacts.TransformCache,
+        artifacts=[standard_artifacts.TransformCache()])
```
1025KB, line 136: add a # TODO(b/162110942): make cache output optional.

tfx/components/transform/executor.py
18jeffreyma: its an optional input artifact, so wasnt sure whether ot include it. will add

18jeffreyma:
Addressed above comments, thanks!

tfx/types/standard_component_specs.py
```diff
@@ -351,12 +351,16 @@ class TransformSpec(ComponentSpec):
...
+      'cache_input_path': ChannelParameter(
+          type=standard_artifacts.TransformCache, optional=True),
   }
   OUTPUTS = {
       'transform_graph':
           ChannelParameter(type=standard_artifacts.TransformGraph),
       'transformed_examples':
           ChannelParameter(type=standard_artifacts.Examples),
+      'cache_output_path':
+          ChannelParameter(type=standard_artifacts.TransformCache),
```
zoyahav, line 363: I thought it is possible now to have optional outputs?
See: https://github.com/tensorflow/tfx/blob/deb618730dc7675c8e9dc75e03b8ca795d49653d/tfx/types/standard_component_specs.py#L361

tfx/components/transform/executor.py
```diff
@@ -279,11 +279,15 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
           should contain two splits 'train' and 'eval'.
         - schema: A list of type `standard_artifacts.Schema` which should
           contain a single schema artifact.
+        - cache_output_path: Optinal cache input of 'tf.Transform', where cached
```
zoyahav, line 282: Would it be possible to resolve this TODO in this change?
https://github.com/tensorflow/tfx/blob/deb618730dc7675c8e9dc75e03b8ca795d49653d/tfx/components/transform/executor.py#L124-L129

This is likely needed in order to enable cache in OSS

zoyahav, line 282: Is it possible to handle this TODO now as well?
https://github.com/tensorflow/tfx/blob/deb618730dc7675c8e9dc75e03b8ca795d49653d/tfx/components/transform/executor.py#L1352
This would likely be very inefficient.

tfx/types/standard_component_specs.py
18jeffreyma: Just saw that PR as well and updated PR too have `cache_output_path` similar to `transformed_examples`.

tfx/components/transform/executor.py
18jeffreyma: DatsetKey now uses a SHA-256 hash of the file_pattern to generate.

1025KB: Hi, Zohar, could you be more specific about this, do you mean we shouldn't copy this?

btw, we haven't enable rolling range for tfx oss transform (next PR, we just enabled multiple output artifact in this [pr](https://github.com/tensorflow/tfx/pull/2160))

18jeffreyma:
Addressed all but most recent comment. Notable changes:

- TransformCache output is now optional. Can be specified using `materialize_cache` boolean. Test coverage added for this.
- DatasetKey is now generated using SHA-256 hashing to reduce collisions.

tfx/components/transform/executor.py
zoyahav: No, I meant that maybe we can see if there's a more performant way of copying on platforms where TFX runs that we can take advantage (for example, snapshot) instead of the naive implementation that is currently used (io_utils.copy_dir).
We can also try to optimize this later when we actually have rolling ranges and we can see being a bottleneck.

tfx/components/transform/executor.py
```diff
@@ -146,8 +141,7 @@ def __init__(self, file_pattern: Text,
       materialize_output_path: The file path where to write the dataset.
     """
     self._file_pattern = file_pattern
-    file_pattern_suffix = os.path.join(
-        *file_pattern.split(os.sep)[-self._FILE_PATTERN_SUFFIX_LENGTH:])
+    file_pattern_suffix = hashlib.sha256(file_pattern.encode()).hexdigest()
```
zoyahav, line 144: Would it be possible to prefix it with the file_pattern_suffix that was used before?
Also, let's rename this variable to dataset_identifier now.

18jeffreyma: Updated. Now uses file_patter_suffix as prefix, with sha256 hash suffix.

18jeffreyma:
Thanks for latest round of feedback! Changed DatasetKey to now also include old file_pattern_suffix

tfx/components/transform/executor.py
```diff
@@ -290,11 +288,15 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
           should contain two splits 'train' and 'eval'.
         - schema: A list of type `standard_artifacts.Schema` which should
           contain a single schema artifact.
+        - cache_output_path: Cache input of 'tf.Transform', where cached
```
zoyahav, line 291: input?

tfx/components/transform/component.py
```diff
@@ -74,8 +74,11 @@ def __init__(
       transform_graph: Optional[types.Channel] = None,
       transformed_examples: Optional[types.Channel] = None,
       input_data: Optional[types.Channel] = None,
+      cache_input_path: Optional[types.Channel] = None,
+      cache_output_path: Optional[types.Channel] = None,
       instance_name: Optional[Text] = None,
       materialize: bool = True,
+      materialize_cache: bool = True,
```
zoyahav, line 81: How about `store_cache` instead of materialize? I think that store makes more sense in the context of cache

tfx/components/transform/executor_test.py
```diff
@@ -227,51 +243,34 @@ def testCounters(self):
...
-    output_cache_artifact.uri = os.path.join(self._output_data_dir, 'CACHE')
-
-    self._output_dict['cache_output_path'] = [output_cache_artifact]
-
     self._exec_properties['module_file'] = self._module_file
     self._transform_executor.Do(self._input_dict, self._output_dict,
                                 self._exec_properties)
     self._verify_transform_outputs()
```
zoyahav, line 250: Should this pass materialize_cache=True?
Same for L265 below

tfx/components/transform/executor_test.py
```diff
@@ -227,51 +243,34 @@ def testCounters(self):
...
+    input_cache_artifact.uri = self._output_cache_artifact.uri
 
     self._make_base_do_params(self._source_data_dir, self._output_data_dir)
 
-    self._input_dict['cache_input_path'] = [input_cache_artifact]
-    self._output_dict['cache_output_path'] = [output_cache_artifact]
+    self._input_dict[labels.CACHE_INPUT_PATH_LABEL] = [input_cache_artifact]
 
     self._exec_properties['module_file'] = self._module_file
     self._transform_executor.Do(self._input_dict, self._output_dict,
```
zoyahav, line 262: Consider using self._runPipelineGetMetrics here instead. This way we can verify that this time the num_instances is 0 (see L223 for example), which means that the input cache was used correctly, and it covers the entire input datasets.

tfx/components/transform/executor.py
18jeffreyma: apologies, typo, fixed

tfx/components/transform/component.py
18jeffreyma: updated, thanks!

tfx/components/transform/executor_test.py
18jeffreyma: by default store_cache is true in `_verify_transform_outputs()`, so no need to specify it unless we want to be more explicit.

tfx/components/transform/executor_test.py
18jeffreyma: since the cache is an analyzer cache, wouldnt the transformed paths still contribute to the num_instances count? my understanding is that the analyzer cache just makes it so we don't need to reanalyze already processed paths, but we still are retransforming everything.

tfx/components/transform/executor_test.py
18jeffreyma: added the check cache tag to be explicit

tfx/components/transform/executor_test.py
18jeffreyma: using metrics, updated test to check that `num_instances` is smaller than initial run `num_instances` by the number of train paths (which should hit cache).

18jeffreyma:
Resolved, latest round of feedback thanks!

tfx/components/transform/executor.py
18jeffreyma: will revisit this once I've submitted the PR for rolling ranges and looked at the performance of this.

PR #2212, opened 2020-07-23
Support Custom Split Config for Transform Component

tinally:
# Support custom split config for Transform component by creating transform.proto and adding splits_config param.


tfx/components/transform/component_test.py
```diff
@@ -92,6 +93,18 @@ def testConstructDuplicateUserModule(self):
...
+                                               transform_splits=['eval'])
+    module_file = '/path/to/preprocessing.py'
+    transform = component.Transform(
+        examples=self.examples,
+        schema=self.schema,
+        module_file=module_file,
+        splits_config=splits_config,
+    )
+    self._verify_outputs(transform)
+
```
1025KB, line 107: verify the execution propeties?

tfx/components/transform/executor.py
```diff
@@ -276,29 +278,45 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+    splits_config = transform_pb2.SplitsConfig()
+    if exec_properties.get('splits_config', None):
+      json_format.Parse(exec_properties['splits_config'], splits_config)
+
+    if not splits_config.analyze_splits:
+      splits_config.analyze_splits.append('train')
+      logging.info("Analyze the 'train' split when splits_config.analyze_splits"
+                   " is not set.")
+    if not splits_config.transform_splits:
+      splits_config.transform_splits.extend(['train', 'eval'])
```
1025KB, line 315: should we extend all splits, not just train and eval?

actually I think the default behavior should be

if exec_properties['splits_config'] is None,
we create a SplitsConfig with analyze `train` and transform all split

if splits_config is set, we don't change what user set

tfx/components/transform/executor.py
```diff
@@ -276,29 +278,45 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
       None
     """
     self._log_startup(input_dict, output_dict, exec_properties)
 
-    train_data_uri = artifact_utils.get_split_uri(input_dict[EXAMPLES_KEY],
-                                                  'train')
-    eval_data_uri = artifact_utils.get_split_uri(input_dict[EXAMPLES_KEY],
-                                                 'eval')
+    splits_config = transform_pb2.SplitsConfig()
+    if exec_properties.get('splits_config', None):
```
1025KB, line 307: what happened if we try to parse None?

tfx/components/transform/executor.py
```diff
@@ -276,29 +278,45 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+        - splits_config: A transform_pb2.SplitsConfig instance, providing splits
+          that should be analyzed and splits that should be transformed. Default
+          behavior is analyze the 'train' split (when analyze_splits is not set)
+          and transform both 'train' and 'eval' splits (when transform_splits is
+          not set).
 
     Returns:
       None
     """
     self._log_startup(input_dict, output_dict, exec_properties)
```
1025KB, line 304: sync to head, I submitted a change

tfx/components/transform/executor.py
```diff
@@ -308,13 +326,26 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
-        output_dict[TRANSFORMED_EXAMPLES_KEY], 'eval')
+        splits_config.transform_splits)
+
+    analyze = []
+    for split in splits_config.analyze_splits:
+      data_uri = artifact_utils.get_split_uri(input_dict[EXAMPLES_KEY], split)
+      analyze.append(io_utils.all_files_pattern(data_uri))
+
+    transform = []
+    transform_output_paths = []
```
1025KB, line 336: we normally use transform_output as transform_graph
here is transformed_examples (materialization)

tfx/components/transform/executor_test.py
```diff
@@ -166,6 +169,28 @@ def testDoWithDuplicatePreprocessingFn(self):
...
 
+  def testDoWithCustomSplits(self):
+    self._exec_properties['splits_config'] = json_format.MessageToJson(
+        transform_pb2.SplitsConfig(analyze_splits=['train'], 
+                                   transform_splits=['eval']),
+        preserving_proto_field_name=True)
+    self._exec_properties['module_file'] = self._module_file
+    self._transformed_examples.split_names = artifact_utils.encode_split_names(
+        ['eval'])
+    self._output_dict[executor.TRANSFORMED_EXAMPLES_KEY] = [self._transformed_examples]
```
1025KB: format?

tfx/proto/transform.proto
```diff
@@ -0,0 +1,27 @@
...
+package tfx.components.transform;
+
+// Defines the splits config.
+message SplitsConfig {
+  // Splits that should be analyzed and used to produce the metadata.
+  // hint: this is normally the train dataset.
+  repeated string analyze_splits = 1;
+
+  // Splits that should be transformed.
+  // hint: this is normally the eval dataset.
```
1025KB, line 25: all split, not just eval

tfx/components/transform/executor.py
tinally: I parsed None in the setUp method in executor_test. If it's None, it will be the default behavior because splits_config.analyze_splits will be empty.

tfx/components/transform/executor.py
tinally: i already synced to head in my last commit and merged them. is anything missing?

tfx/components/transform/executor_test.py
tinally: you mean pylint format issues? I added another commit to fix all the pylint issues. this is an older version.

tfx/components/transform/executor.py
tinally: yes, I think it makes sense to transform all splits, I will make this change

tinally: if splits_config is set empty by the user, I think we still want to change it to the default behavior. that's why i use if analyze_splits is empty to check for whether we want default behavior.

tfx/components/transform/executor.py
tinally: ah i understand, thanks for the catch

tfx/components/transform/executor.py
```diff
@@ -276,45 +277,75 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+    transform_example_artifact.split_names = artifact_utils.encode_split_names(
+        splits_config.transform_splits)
+
+    analyze = []
+    for split in splits_config.analyze_splits:
+      data_uri = artifact_utils.get_split_uri(input_dict[EXAMPLES_KEY], split)
+      analyze.append(io_utils.all_files_pattern(data_uri))
+
+    transform = []
+    transformed_examples = []
```
1025KB, line 337: align with TRANSFORM_MATERIALIZE_OUTPUT_PATHS?

check with Jeffrey, he is going to update transform to support mutiple artifacts, so he probably will update the name too

tfx/components/transform/executor.py
```diff
@@ -652,7 +679,7 @@ def to_runner_api_pickled(self, context):
       # picklable.
       return self.to_runner_api_parameter(context)
 
-    def expand(
+    def expand( # pylint: disable=arguments-differ
```
1025KB, line 682: two space before, let's not add it, I feel internal doesn't care about this lint

tfx/components/transform/executor.py
```diff
@@ -536,6 +558,7 @@ class _ToArrowRecordBatchesFn(beam.DoFn):
     """Converts a batch of serialized examples to an Arrow RecordBatch."""
 
     def __init__(self, schema: Optional[schema_pb2.Schema]):
+      super().__init__()
```
1025KB, line 561: is there an issue before adding those? let's keep the change minimum otherwise there will be a lot of conflict. refactor can be done in separate PRs if needed

tfx/components/transform/executor.py
18jeffreyma: my first draft of proposed executor changes are here, for reference:

https://github.com/18jeffreyma/tfx/commit/5ab8e5975371c9537f50058cce6746a47432f301

tfx/components/transform/executor.py
tinally: because it's possible user just set analyze but not transform, so i checked splits_config.analyze_splits and splits_config.transform_splits separately. if we directly check for splits_config, if splits_config is not set correctly (maybe user sets analyze or transform to be empty), then i don't know which one i should add the default behavior.

tfx/components/transform/executor.py
tinally: this case shouldn't happen? we only reach json_format.Parse after line 308 checks that 'splits_config' is not None.

tfx/components/transform/executor.py
```diff
@@ -276,45 +277,75 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+    for split in splits_config.analyze_splits:
+      data_uri = artifact_utils.get_split_uri(input_dict[EXAMPLES_KEY], split)
+      analyze.append(io_utils.all_files_pattern(data_uri))
+
+    transform = []
+    transformed_examples = []
+    for split in splits_config.transform_splits:
+      data_uri = artifact_utils.get_split_uri(input_dict[EXAMPLES_KEY], split)
+      transform.append(io_utils.all_files_pattern(data_uri))
+      transformed_output = artifact_utils.get_split_uri(
```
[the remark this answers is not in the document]
tinally: line 341, this was originally named 'transformed_train_output' and 'transformed_eval_output'.

i changed line 343 to transformed_examples.

tfx/components/transform/executor.py
tinally: yes, the pylint test failed before I added this, the same pylint issue like tfdv

tfx/components/transform/executor.py
tinally: got it, I will check it, thanks!!

tfx/components/transform/executor.py
tinally: ok i removed this pylint

tfx/components/transform/executor_test.py
```diff
@@ -166,10 +169,32 @@ def testDoWithDuplicatePreprocessingFn(self):
...
+    path_to_saved_model = os.path.join(
+        self._transformed_output.uri, tft.TFTransformOutput.TRANSFORM_FN_DIR,
+        tf.saved_model.SAVED_MODEL_FILENAME_PB)
+    self.assertTrue(tf.io.gfile.exists(path_to_saved_model))
+
   def testCounters(self):
     self._exec_properties['preprocessing_fn'] = self._preprocessing_fn
-    metrics = self._runPipelineGetMetrics(self._input_dict, self._output_dict,
-                                          self._exec_properties)
+    metrics = self._runPipelineGetMetrics()
```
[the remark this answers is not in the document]
tinally: yes, i double checked, _runPipelineGetMetrics have unused arguments, these fields are still accessed with self in the method, so i removed these arguments

tfx/components/transform/executor.py
```diff
@@ -287,29 +289,46 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+    if exec_properties.get('splits_config', None):
+      json_format.Parse(exec_properties['splits_config'], splits_config)
+      if not splits_config.analyze_splits:
+        raise ValueError(
+            'analyze_splits cannot be empty when splits_config is set.')
+    else:
+      splits_config.analyze_splits.append('train')
+      splits_config.transform_splits.extend(
+          artifact_utils.decode_split_names(
+              input_dict[EXAMPLES_KEY][0].split_names))
```
1025KB, line 328: use artifact_utils.get_single_instance here

tfx/components/transform/executor.py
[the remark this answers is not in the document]
1025KB: don't use transformed_output here transformed_output normally means transformed_graph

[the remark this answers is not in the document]
tinally: oh i see, ok

tfx/components/transform/executor.py
1025KB: let's drop this, internal lint doesn't check this

tfx/components/transform/executor_test.py
```diff
@@ -178,10 +181,42 @@ def testDoWithDuplicatePreprocessingFn(self):
...
+    self._output_dict[executor.TRANSFORMED_EXAMPLES_KEY] = [
+        self._transformed_examples]
+
+    self._transform_executor.Do(self._input_dict, self._output_dict,
+                                self._exec_properties)
+    self.assertNotEqual(
+        0,
+        len(
+            tf.io.gfile.listdir(
+                os.path.join(self._transformed_examples.uri, 'eval'))))
```
1025KB, line 201: check train is not generated

tinally: got it, it's added now

tfx/components/transform/component.py
```diff
@@ -108,12 +111,17 @@ def preprocessing_fn(inputs: Dict[Text, Any], custom_config:
...
+        behavior (when splits_config is not set) is analyze the 'train' split
+        and transform all splits. If splits_config is set, analyze_splits cannot
+        be empty.
       transform_graph: Optional output 'TransformPath' channel for output of
         'tf.Transform', which includes an exported Tensorflow graph suitable for
         both training and serving;
       transformed_examples: Optional output 'ExamplesPath' channel for
         materialized transformed examples, which includes both 'train' and
-        'eval' splits.
+        'eval' splits if custom splits is not specified in splits_config.
```
zoyahav, line 124: Should this mention what it should include if splits_config is specified? (similarly above for examples)

tfx/components/transform/executor.py
```diff
@@ -287,29 +289,48 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
     Args:
       input_dict: Input dict from input key to a list of artifacts, including:
         - input_data: A list of type `standard_artifacts.Examples` which
-          should contain two splits 'train' and 'eval'.
+          should contain two splits 'train' and 'eval' if custom splits
```
zoyahav, line 292: Same here and below, and if it is?

tfx/components/transform/executor.py
```diff
@@ -287,29 +289,48 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+          includes both 'train' and 'eval' splits if custom splits is not
+          specified in splits_config.
+      exec_properties: A dict of execution properties, including:
         - module_file: The file path to a python module file, from which the
           'preprocessing_fn' function will be loaded.
         - preprocessing_fn: The module path to a python function that
-          implements 'preprocessing_fn'.
+          implements 'preprocessing_fn'. Exactly one of 'module_file' and
+          'preprocessing_fn' should be set.
+        - splits_config: A transform_pb2.SplitsConfig instance, providing splits
```
zoyahav, line 308: May want to explicitly mention that it's okay if there's an overlap between the splits.

tfx/components/transform/executor.py
```diff
@@ -356,16 +382,13 @@ def _GetCachePath(label, params_dict):
...
         labels.ANALYZE_PATHS_FILE_FORMATS_LABEL:
-            labels.FORMAT_TFRECORD,
-        labels.TRANSFORM_DATA_PATHS_LABEL: [
-            io_utils.all_files_pattern(train_data_uri),
-            io_utils.all_files_pattern(eval_data_uri)
-        ],
-        labels.TRANSFORM_PATHS_FILE_FORMATS_LABEL: [
-            labels.FORMAT_TFRECORD, labels.FORMAT_TFRECORD
-        ],
+            [labels.FORMAT_TFRECORD for _ in range(len(analyze_data_paths))],
```
zoyahav, line 387: This can be simplified with 
`[labels.FORMAT_TFRECORD] * len(analyze_data_paths)`

tfx/components/transform/executor.py
```diff
@@ -1195,7 +1220,7 @@ def _ExtractRawExampleBatches(record_batch):
   def _RunInPlaceImpl(
       self, preprocessing_fn: Any,
       metadata: dataset_metadata.DatasetMetadata,
-      typespecs: Dict[Text, tf.TypeSpec],
+      typespecs: Dict[Text, tf.TypeSpec], # pylint: disable=unused-argument
```
zoyahav, line 1223: Can we just remove it then?

tfx/components/transform/executor_test.py
```diff
@@ -129,7 +132,7 @@ def _verify_transform_outputs(self, materialize=True):
         tf.saved_model.SAVED_MODEL_FILENAME_PB)
     self.assertTrue(tf.io.gfile.exists(path_to_saved_model))
 
-  def _runPipelineGetMetrics(self, inputs, outputs, exec_properties):
+  def _run_pipeline_get_metrics(self):
```
zoyahav, line 135: Most of this file is camelCase, so I suggest switching those that are not to camelCase instead for consistency.

tfx/components/transform/executor_test.py
```diff
@@ -191,10 +194,45 @@ def testDoWithDuplicatePreprocessingFn(self):
...
+        transform_pb2.SplitsConfig(analyze_splits=['train'],
+                                   transform_splits=['eval']),
+        preserving_proto_field_name=True)
+    self._exec_properties['module_file'] = self._module_file
+    self._transformed_examples.split_names = artifact_utils.encode_split_names(
+        ['eval'])
+    self._output_dict[executor.TRANSFORMED_EXAMPLES_KEY] = [
+        self._transformed_examples]
+
+    self._transform_executor.Do(self._input_dict, self._output_dict,
```
zoyahav, line 208: Is there a way to test the the contents of these splits is as expected?
We want to make sure that the right data ends up in the right splits.

tfx/proto/transform.proto
```diff
@@ -0,0 +1,27 @@
...
+// limitations under the License.
+syntax = "proto3";
+
+package tfx.components.transform;
+
+// Defines the splits config.
+message SplitsConfig {
+  // Splits that should be analyzed and used to produce the metadata.
+  // hint: this is normally the train dataset.
+  repeated string analyze_splits = 1;
```
zoyahav, line 22: Any reason why this shouldn't be analyze_and_transform? Do we expect a case where a split is only used for analysis and it shouldn't be transformed?

tfx/components/transform/executor_test.py
```diff
@@ -191,10 +194,45 @@ def testDoWithDuplicatePreprocessingFn(self):
...
+        0,
+        len(
+            tf.io.gfile.listdir(
+                os.path.join(self._transformed_examples.uri, 'eval'))))
+    path_to_saved_model = os.path.join(
+        self._transformed_output.uri, tft.TFTransformOutput.TRANSFORM_FN_DIR,
+        tf.saved_model.SAVED_MODEL_FILENAME_PB)
+    self.assertTrue(tf.io.gfile.exists(path_to_saved_model))
+
+  def testDoWithEmptyAnalyzeSplits(self):
```
zoyahav, line 223: Should we also have a test case of empty transform_splits?

tinally: i think empty transform_splits is not a special case, it's equivalent to what is tested in testDoWithCustomSplits

i can still add a test for empty transform_splits if you think it's necessary and more clear

tfx/components/transform/executor.py
tinally: yes, i will remove this then

tfx/proto/transform.proto
tinally: i used analyze_splits and transform_splits to correspond to labels.ANALYZE_DATA_PATHS_LABEL and labels.TRANSFORM_DATA_PATHS_LABEL in the executor.

if we use analyze_and_transform, then we need to do additional work to combine analyze_and_transform and transform_splits for the labels.TRANSFORM_DATA_PATHS_LABEL.

tfx/components/transform/component.py
tinally: i see, i updated the comments, thanks!

tinally: i see, i updated the comments, thanks!

tfx/components/transform/executor_test.py
tinally: I see, I will follow the test in ExampleGen to check their sizes.

tfx/proto/transform.proto
tinally: @rose-rong-liu

tfx/components/transform/executor_test.py
tinally: The reason why I updated the name for this was because it originally had a pylint error -- invalid-name.

I don't know if we should switch those to camel case because I checked other executor_test files and I found out that helper functions are usually like this (_verify_test) and only tests are camel case (testDo). You think it's ok to leave it like this and be consistent with other files?

tfx/proto/transform.proto
zoyahav: I think this is a question of API as opposed to component logic, since the alternative logic would have the same level of complexity, and this is what we currently do in TFX.
So I think that the API question is, do we want users to explicitly specify analyze and transform splits separately? Are there any cases where the user would like to analyze a split, but not transform it (if they're transforming anything)?

```
  splits_config {
    analyze_splits: "train1"
    analyze_splits: "train2"
    transform_splits: "train1"
    transform_splits: "train2"
    transform_splits: "eval"
  }
```
vs.
```
  splits_config {
    analyze_and_transform: "train1"
    analyze_and_transform: "train2"
    transform_splits: "eval"
  }
```

I'd be fine with either option, just want to make sure that we're making a conscious choice here (@rose-rong-liu feel free to chime in).

Also, could we rename analyze_splits/transform_splits->analyze/transform? I think that would be cleaner.

rose-rong-liu: In fact we started with analyze_and_transform initially.
Then chenkai's cl/266415037 (columnar support) separated it up, which also changes the implementation of internal transform/runner.py.
I think either way works. If we switch it back, we also need to change transform/runner.py.

1025KB: another thought is if we turn off materialization for transformed example, then set up analyze_and_transform might be a little bit weird.

Re Rose, for the label I think keep then separate is fine, just for the config:
-  analyze_and_transform/transform means the config is shorter as Zohar's example above.
-  analyze/transform is more accurate if materialization is off.

Either is fine to me.

tinally: I renamed analyze_splits/transform_splits to analyze/transform.

tfx/components/transform/executor_test.py
tinally: Yes, I can do this, so to confirm, I will change all the method names to snake_case, including the tests?

tinally: Hi Zohar, I updated all the tft tests to snake_case.

tfx/proto/transform.proto
zoyahav: Keeping analyze/transform is fine by me.

tfx/components/transform/executor_test.py
zoyahav: Yes, thank you.

tfx/components/transform/executor_test.py
zoyahav: Sg, was this done?

zoyahav (approved):
Thanks!

tfx/components/transform/executor_test.py
tinally: Yes, I added assertGreater to compare the transformed examples for 'train' and 'eval'.

tfx/proto/transform.proto
tinally: Great, thank you!

PR #2270, opened 2020-08-03
Move artifact setup to executor for all components

tinally:
# Move artifact setup to executor instead of in component constructor for all components.


tfx/components/example_gen/base_example_gen_executor.py
```diff
@@ -281,6 +281,15 @@ def Do(
     """
     self._log_startup(input_dict, output_dict, exec_properties)
 
+    input_config = example_gen_pb2.Input()
+    output_config = example_gen_pb2.Output()
+    json_format.Parse(exec_properties[utils.INPUT_CONFIG_KEY], input_config)
+    json_format.Parse(exec_properties[utils.OUTPUT_CONFIG_KEY], output_config)
+
+    for output_examples_artifact in output_dict[utils.EXAMPLES_KEY]:
```
1025KB, line 289: use artifact_utils.get_single_instance (no need to have for loop, get_single_instance will assert it only contains one), same else where

tinally: got it, i updated it for other components as well

tfx/components/example_gen/component_test.py
```diff
@@ -129,9 +119,6 @@ def testConstructWithOutputConfig(self):
                      example_gen.outputs['examples'].type_name)
     artifact_collection = example_gen.outputs['examples'].get()
     self.assertEqual(1, len(artifact_collection))
-    self.assertEqual(['train', 'eval', 'test'],
```
1025KB, line 132: you can verify 
example_gen.exec_properties['input_config'] and example_gen.exec_properties['output_config'] to make sure it's set correctly

same for testConstructWithInputConfig

you don't need to add `self.assertEqual(['train', 'eval', xxx], generate_output_split_names(input_config, output_config))` in this test, but make sure example_gen/utils_test.py covers such input and output config combination for generate_output_split_names

tfx/components/statistics_gen/executor.py
```diff
@@ -87,6 +87,24 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+    # Load and deserialize exclude splits from execution properties.
+    exclude_splits = json_utils.loads(
+        exec_properties.get(EXCLUDE_SPLITS_KEY, 'null')) or []
+    if not isinstance(exclude_splits, list):
+      raise ValueError('exclude_splits in execution properties needs to be a '
+                       'list. Got %s instead.' % type(exclude_splits))
+
+    examples = artifact_utils.get_single_instance(input_dict[EXAMPLES_KEY])
+    statistics_artifact = artifact_utils.get_single_instance(
+        output_dict[STATISTICS_KEY])
```
1025KB, line 99: you can update line 139 to use this stats_artifact instead of reaccess from output dict, same eleswhere

tfx/components/example_gen/component_test.py
tinally: i see, i updated the component_test

also i checked -- utils_test.py will check for generate_output_split_names

tfx/components/example_gen/base_example_gen_executor.py
```diff
@@ -289,8 +299,7 @@ def Do(
       for split_name, example_split in example_splits.items():
         (example_split
          | 'WriteSplit[{}]'.format(split_name) >> _WriteSplit(
-             artifact_utils.get_split_uri(output_dict[utils.EXAMPLES_KEY],
-                                          split_name)))
+             artifact_utils.get_split_uri([examples_artifact], split_name)))
```
1025KB, line 302: ah my bad, I thought we could use examples_artifact directly, if not, just keep it as it is (revert the change)

tfx/components/example_gen/component_test.py
```diff
@@ -112,42 +105,45 @@ def testConstructCustomExecutor(self):
...
-                     artifact_utils.decode_split_names(
-                         artifact_collection[0].split_names))
+
+    stored_output_config = example_gen_pb2.Output()
+    json_format.Parse(example_gen.exec_properties['output_config'],
+                      stored_output_config)
+    self.assertEqual(output_config, stored_output_config)
```
1025KB, line 127: test if input_config is set to default?

tfx/components/example_gen/component_test.py
```diff
@@ -112,42 +105,45 @@ def testConstructCustomExecutor(self):
...
     artifact_collection = example_gen.outputs['examples'].get()
     self.assertEqual(1, len(artifact_collection))
-    self.assertEqual(['train', 'eval', 'test'],
-                     artifact_utils.decode_split_names(
-                         artifact_collection[0].split_names))
+
+    stored_input_config = example_gen_pb2.Input()
+    json_format.Parse(example_gen.exec_properties['input_config'],
+                      stored_input_config)
+    self.assertEqual(input_config, stored_input_config)
```
1025KB, line 146: test if output config is set to default

tfx/components/example_gen/base_example_gen_executor.py
tinally: no worries, i changed it back

tfx/components/example_gen/component.py
```diff
@@ -93,10 +93,8 @@ def __init__(self,
     output_config = output_config or utils.make_default_output_config(
         input_config)
     if not example_artifacts:
-      artifact = standard_artifacts.Examples()
-      artifact.split_names = artifact_utils.encode_split_names(
-          utils.generate_output_split_names(input_config, output_config))
-      example_artifacts = channel_utils.as_channel([artifact])
+      example_artifacts = channel_utils.as_channel(
+          [standard_artifacts.Examples()])
```
ruoyu90, line 97: @1025KB I vaguely remember there is a reason that we cannot drop the artifact yet. Is it for cloud?

/cc @numerology

PR #2274, opened 2020-08-04
Added abstract BeamExecutor class

kshivvy:
The abstract class BeamExecutor will be implemented by TFX components that utilize Beam pipelines (e.g. ExampleGen, StatisticsGen), for future Beam Fusion optimization.

### DIFF
--- a/tfx/components/base/base_executor.py
+++ b/tfx/components/base/base_executor.py
@@ -22,7 +22,7 @@ import abc
 import json
 import multiprocessing
 import os
-from typing import Any, Dict, List, Optional, Text
+from typing import Any, Dict, List, Optional, Text, Union, Tuple
 
 import absl
 import apache_beam as beam
@@ -75,7 +75,7 @@ class BaseExecutor(with_metaclass(abc.ABCMeta, object)):
       exec_properties: A dict of execution properties. These are inputs to
         pipeline with primitive types (int, string, float) and fully
         materialized when a pipeline is constructed. No dependency to other
-        component or later injection from orchestration systems is necessary or
+        components or later injection from orchestration systems is necessary or
         possible on these values.
 
     Returns:
@@ -162,14 +162,16 @@ class EmptyExecutor(BaseExecutor):
     pass
 
 
-class BeamExecutor(BaseExecutor):
-  """ Abstract TFX executor class for handling Beam-related logic."""
+class FuseableBeamExecutor(BaseExecutor):
+  """ Abstract TFX executor class for handling Beam-related logic.
+      Experimental: no backwards compatability guarantees.
+  """
 
   @abc.abstractmethod
   def beam_io_signature(self, input_dict: Dict[Text, List[types.Artifact]],
                         output_dict: Dict[Text, List[types.Artifact]],
-                        exec_properties: Dict[Text, Any]
-                        ) -> List[Dict[Text, Any]]:
+                        exec_properties: Dict[Text, Union[int, float, Text]]
+                        ) -> Tuple[Dict[Text, type], Dict[Text, type]]:
     """Provides input and output signatures for the input_dict and output_dict
        of the underlying component.
 
@@ -182,7 +184,7 @@ class BeamExecutor(BaseExecutor):
       exec_properties: A dict of execution properties. These are inputs to
         pipeline with primitive types (int, string, float) and fully
         materialized when a pipeline is constructed. No dependency to other
-        component or later injection from orchestration systems is necessary or
+        components or later injection from orchestration systems is necessary or
         possible on these values.
 
     Returns:
@@ -196,7 +198,7 @@ class BeamExecutor(BaseExecutor):
   def read_inputs(self, pipeline: beam.Pipeline,
                   input_dict: Dict[Text, List[types.Artifact]],
                   output_dict: Dict[Text, List[types.Artifact]],
-                  exec_properties: Dict[Text, Any]
+                  exec_properties: Dict[Text, Union[int, float, Text]]
                   ) -> Dict[Text, beam.pvalue.PCollection]:
     """Executes reads for underlying component implementation.
 
@@ -209,7 +211,7 @@ class BeamExecutor(BaseExecutor):
       exec_properties: A dict of execution properties. These are inputs to
         pipeline with primitive types (int, string, float) and fully
         materialized when a pipeline is constructed. No dependency to other
-        component or later injection from orchestration systems is necessary or
+        components or later injection from orchestration systems is necessary or
         possible on these values.
 
     Returns:
@@ -219,12 +221,12 @@ class BeamExecutor(BaseExecutor):
     pass
 
   @abc.abstractmethod
-  def construct_beam_graph(self, pipeline: beam.Pipeline,
-                           beam_inputs: Dict[Text, beam.pvalue.PCollection],
-                           input_dict: Dict[Text, List[types.Artifact]],
-                           output_dict: Dict[Text, List[types.Artifact]],
-                           exec_properties: Dict[Text, Any]
-                           ) -> Dict[Text, beam.pvalue.PCollection]:
+  def run_component(self, pipeline: beam.Pipeline,
+                    beam_inputs: Dict[Text, beam.pvalue.PCollection],
+                    input_dict: Dict[Text, List[types.Artifact]],
+                    output_dict: Dict[Text, List[types.Artifact]],
+                    exec_properties: Dict[Text, Union[int, float, Text]]
+                    ) -> Dict[Text, beam.pvalue.PCollection]:
     """Executes underlying component implementation, except for reads/writes.
 
     Args:
@@ -238,7 +240,7 @@ class BeamExecutor(BaseExecutor):
       exec_properties: A dict of execution properties. These are inputs to
         pipeline with primitive types (int, string, float) and fully
         materialized when a pipeline is constructed. No dependency to other
-        component or later injection from orchestration systems is necessary or
+        components or later injection from orchestration systems is necessary or
         possible on these values.
 
     Returns:
@@ -252,12 +254,13 @@ class BeamExecutor(BaseExecutor):
                     beam_outputs: Dict[Text, beam.pvalue.PCollection],
                     input_dict: Dict[Text, List[types.Artifact]],
                     output_dict: Dict[Text, List[types.Artifact]],
-                    exec_properties: Dict[Text, Any]) -> None:
+                    exec_properties: Dict[Text, Union[int, float, Text]]
+                    ) -> None:
     """Executes writes for underlying component implementation.
 
     Args:
       beam_outputs: A Beam output dict from input key to PCollections. Output
-        of self.construct_beam_graph().
+        of self.run_component().
       input_dict: Input dict from input key to a list of Artifacts. These are
         often outputs of another component in the pipeline and passed to the
         component by the orchestration system.
@@ -266,7 +269,7 @@ class BeamExecutor(BaseExecutor):
       exec_properties: A dict of execution properties. These are inputs to
         pipeline with primitive types (int, string, float) and fully
         materialized when a pipeline is constructed. No dependency to other
-        component or later injection from orchestration systems is necessary or
+        components or later injection from orchestration systems is necessary or
         possible on these values.
 
     Returns:
--- a/tfx/components/base/base_executor.py
+++ b/tfx/components/base/base_executor.py
@@ -163,13 +163,33 @@ class EmptyExecutor(BaseExecutor):
 
 
 class BeamExecutor(BaseExecutor):
-  """ TFX executor class for handling Beam-related logic."""
+  """ Abstract TFX executor class for handling Beam-related logic."""
 
   @abc.abstractmethod
   def beam_io_signature(self, input_dict: Dict[Text, List[types.Artifact]],
                         output_dict: Dict[Text, List[types.Artifact]],
                         exec_properties: Dict[Text, Any]
                         ) -> List[Dict[Text, Any]]:
+    """Provides input and output signatures for the input_dict and output_dict
+       of the underlying component.
+
+    Args:
+      input_dict: Input dict from input key to a list of Artifacts. These are
+        often outputs of another component in the pipeline and passed to the
+        component by the orchestration system.
+      output_dict: Output dict from output key to a list of Artifacts. These are
+        often consumed by a dependent component.
+      exec_properties: A dict of execution properties. These are inputs to
+        pipeline with primitive types (int, string, float) and fully
+        materialized when a pipeline is constructed. No dependency to other
+        component or later injection from orchestration systems is necessary or
+        possible on these values.
+
+    Returns:
+      A first dict from input artifact name/split key to the return type of
+      reading the inputs. A second dict from output artifact name/split key to
+      the return type of executing underlying component logic.
+    """
     pass
 
   @abc.abstractmethod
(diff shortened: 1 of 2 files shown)


kshivvy:
@charlesccychen 
@yifanmai

tfx/components/base/base_executor.py
```diff
@@ -160,3 +160,116 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
          output_dict: Dict[Text, List[types.Artifact]],
          exec_properties: Dict[Text, Any]) -> None:
     pass
+
+
+class BeamExecutor(BaseExecutor):
```
charlesccychen, line 165: Can you rename this `FuseableBeamExecutor`?

CC: @yifanmai who may have an opinion on this name (should we call this `ExperimentalFuseableBeamExecutor`?)

tfx/components/base/base_executor.py
```diff
@@ -160,3 +160,116 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+  """ Abstract TFX executor class for handling Beam-related logic."""
+
```
charlesccychen, line 167: Can you add a line to the effect "Experimental: no backwards compatibility guarantees."?

tfx/components/base/base_executor_test.py
```diff
@@ -37,21 +44,130 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
 
 
+class _BeamTestExecutor(base_executor.BeamExecutor):
+  """Fake BeamExecutor for testing purpose only."""
+
+  def beam_io_signature(self, input_dict: Dict[Text, List[types.Artifact]],
+                        output_dict: Dict[Text, List[types.Artifact]],
+                        exec_properties: Dict[Text, Any]
+                        ) -> List[Dict[Text, Any]]:
+    input_signature = {_EXAMPLES: io.TextIOBase}
```
charlesccychen, line 54: Should this just be `Text`?

tfx/components/base/base_executor_test.py
```diff
@@ -37,21 +44,130 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+    return input_signature, output_signature
+
+  def read_inputs(self, pipeline: beam.Pipeline,
+                  input_dict: Dict[Text, List[types.Artifact]],
+                  output_dict: Dict[Text, List[types.Artifact]],
+                  exec_properties: Dict[Text, Any]
+                  ) -> Dict[Text, beam.pvalue.PCollection]:
+    input_dir = input_dict[_EXAMPLES].uri
+    beam_inputs = ({_EXAMPLES: (pipeline
+                                | beam.io.textio.ReadFromText(input_dir))})
```
charlesccychen, line 65: Can you make this `beam.io.textio.ReadFromText(input_dir).with_output_types(Text)`? (see comment below regarding typing)

tfx/components/base/base_executor_test.py
```diff
@@ -37,21 +44,130 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+        f.write(line)
+
+  def testBeamExecutorBeamIoSignature(self):
+    beam_test_executor = _BeamTestExecutor()
+    input_signature, output_signature = (
+        beam_test_executor.beam_io_signature(
+            self._input_dict, self._output_dict, {}))
+
+    with open(self._input_data_path) as f:
+      self.assertTrue(isinstance(f, input_signature[_EXAMPLES]))
```
charlesccychen, line 157: Instead of opening the file and asserting, can you assert that `self._input_dict[_EXAMPLES]` has the declared input signature (i.e. `Text`)?

tfx/components/base/base_executor_test.py
```diff
@@ -37,21 +44,130 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+
+    with open(self._full_output_data_path) as f:
+      self.assertTrue(isinstance(f, output_signature[_SHUFFLED_EXAMPLES]))
```
charlesccychen, line 160: Similarly, can you assert that `self._output_dict[_EXAMPLES]` has the appropriate output signature (i.e. `int`)?  Note that the system should be able to infer this (otherwise, you would need to use `.with_output_types(int)` during the `construct_beam_graph()` method).

charlesccychen:
CC: @ruoyu90 @zhitaoli

tfx/components/base/base_executor.py
yifanmai: `FuseableBeamExecutor` and `ExperimentalFuseableBeamExecutor` both sound good to me.

tfx/components/base/base_executor.py
```diff
@@ -160,3 +160,116 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+    Args:
+      input_dict: Input dict from input key to a list of Artifacts. These are
+        often outputs of another component in the pipeline and passed to the
+        component by the orchestration system.
+      output_dict: Output dict from output key to a list of Artifacts. These are
+        often consumed by a dependent component.
+      exec_properties: A dict of execution properties. These are inputs to
+        pipeline with primitive types (int, string, float) and fully
+        materialized when a pipeline is constructed. No dependency to other
+        component or later injection from orchestration systems is necessary or
```
yifanmai, line 185: nit: component -> components

tfx/components/base/base_executor.py
```diff
@@ -160,3 +160,116 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+        possible on these values.
+
+    Returns:
```
yifanmai, line 188: If only two dicts are returned, can this be a `Tuple[List[Dict[Text, Any]], List[Dict[Text, Any]]]` instead of a Tuple?

Additionally, should `Any` be `type` instead?

tfx/components/base/base_executor.py
```diff
@@ -160,3 +160,116 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+      A dict from input artifact name/split key to PCollections from reads for
+      underlying component execution logic.
+    """
+    pass
+
+  @abc.abstractmethod
+  def construct_beam_graph(self, pipeline: beam.Pipeline,
```
yifanmai, line 222: @charlesccychen I'm not a fan of this name because it's not parallel with `read_inputs` and `write_outputs` - those two methods also construct Beam graphs, so why should this method have the claim to the name?

Something like `execute_component` or `run_component` or `process_artifacts` might make more sense?

tfx/components/base/base_executor.py
```diff
@@ -160,3 +160,116 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+
+
+class BeamExecutor(BaseExecutor):
+  """ Abstract TFX executor class for handling Beam-related logic."""
+
+  @abc.abstractmethod
+  def beam_io_signature(self, input_dict: Dict[Text, List[types.Artifact]],
+                        output_dict: Dict[Text, List[types.Artifact]],
+                        exec_properties: Dict[Text, Any]
+                        ) -> List[Dict[Text, Any]]:
```
yifanmai, line 172: You can do `from typing import Union`, then the return type can be `List[Dict[Text, Union[int, str, float]]]`.

tfx/components/base/base_executor_test.py
```diff
@@ -37,21 +44,130 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+  def write_outputs(self, pipeline: beam.Pipeline,
+                    beam_outputs: Dict[Text, beam.pvalue.PCollection],
+                    input_dict: Dict[Text, List[types.Artifact]],
+                    output_dict: Dict[Text, List[types.Artifact]],
+                    exec_properties: Dict[Text, Any]) -> None:
+    output_dir = output_dict[_SHUFFLED_EXAMPLES].uri
+    _ = (beam_outputs[_SHUFFLED_EXAMPLES]
+         | beam.io.textio.WriteToText(output_dir))
+
+  def Do(self, input_dict: Dict[Text, List[types.Artifact]],
```
yifanmai, line 89: @charlesccychen Should Do be defined here or on `FuseableBeamExecutor`?

tfx/components/base/base_executor_test.py
```diff
@@ -37,21 +44,130 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+         output_dict: Dict[Text, List[types.Artifact]],
+         exec_properties: Dict[Text, Any]) -> None:
+    with beam.Pipeline(runner=beam.runners.create_runner('DirectRunner')) as p:
```
yifanmai, line 92: Have this take in a Pipeline (or PipelineOptions) from outside instead of configuring the runner in the method.

tfx/components/base/base_executor_test.py
```diff
@@ -37,21 +44,130 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+    self._input_dict = {_EXAMPLES: examples}
+
+    # Create output dict
+    shuffled_examples = types.standard_artifacts.Examples()
+    shuffled_examples.uri = output_data_path
+    self._output_dict = {_SHUFFLED_EXAMPLES: shuffled_examples}
+
+    # Run Do() method
+    beam_test_executor = _BeamTestExecutor()
+    beam_test_executor.Do(self._input_dict, self._output_dict, {})
```
yifanmai, line 142: Move `Do` into the test methods since it won't be used in every method after addressing @charlesccychen's comments.

tfx/components/base/base_executor.py
kshivvy: Done, I changed it to `FuseableBeamExecutor`

tfx/components/base/base_executor.py
kshivvy: Done.

tfx/components/base/base_executor_test.py
kshivvy: You're right, I changed the types to `Text` and `int`

tfx/components/base/base_executor_test.py
kshivvy: Done, I added that `.with_output_types(Text)`

tfx/components/base/base_executor_test.py
kshivvy: I got rid of the file openings, and assert that `input_signature[_EXAMPLES] = Text` and `output_signature[_SHUFFLED_EXAMPLES] = int`

tfx/components/base/base_executor_test.py
kshivvy: I got rid of the file openings, and assert that `input_signature[_EXAMPLES] = Text` and `output_signature[_SHUFFLED_EXAMPLES] = int`. I'm unable to force `.with_output_types(int)` in `construct_beam_graph()` since I get the error `  File "apache_beam/coders/stream.pyx", line 225, in apache_beam.coders.stream.get_varint_size
TypeError: an integer is required [while running 'WriteToText/Write/WriteImpl/FinalizeWrite']`, so I have to do `actual_sum = int(f.read())` in `def testFuseableBeamExecutorDo(self):`. Let me know if this is the correct approach.

tfx/components/base/base_executor.py
kshivvy: Changed. This typo was in a 5 places throughout the file so I modified those as well.

tfx/components/base/base_executor.py
kshivvy: You are correct. I changed it to `Tuple[Dict[Text, type]] Dict[Text, type]]`.

tfx/components/base/base_executor.py
kshivvy: I changed the name to `run_component` to make it parallel. @charlesccychen let me know if this name is alright for you.

tfx/components/base/base_executor.py
kshivvy: Done, I changed the return type as suggested.

tfx/components/base/base_executor_test.py
kshivvy: The `BaseExecutor` class has a `_make_beam_pipeline()` method, which is used in `executor.py` for each component. I modified my `Do()` method to use that pipeline.

tfx/components/base/base_executor_test.py
kshivvy: Done.

tfx/components/base/base_executor_test.py
kshivvy: I believe `Do()` should be defined in the `_BeamTestExecutor` class, since the generic implementation of `read_inputs()`, `run_component()`, `write_outputs()` is not possible in the components without extreme refactoring

tfx/components/base/base_executor.py
```diff
@@ -160,3 +160,119 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+    """
+    pass
+
+  @abc.abstractmethod
+  def write_outputs(self, pipeline: beam.Pipeline,
+                    beam_outputs: Dict[Text, beam.pvalue.PCollection],
+                    input_dict: Dict[Text, List[types.Artifact]],
+                    output_dict: Dict[Text, List[types.Artifact]],
+                    exec_properties: Dict[Text, Union[int, float, Text]]
+                    ) -> None:
```
yifanmai, line 258: Question for @charlesccychen: In Beam Python, sink PTransforms typically return PCollections that represent when the write has been completed; should those PCollections be exposed in the return value here?

Don't need to change this in this PR; just a thought.

tfx/components/base/base_executor_test.py
```diff
@@ -18,40 +18,153 @@
...
+import os
+from typing import Dict, List, Text, Union, Tuple
 
+import apache_beam as beam
 from apache_beam.options.pipeline_options import DirectOptions
 from apache_beam.options.pipeline_options import StandardOptions
 import tensorflow as tf
 
 from tfx import types
 from tfx.components.base import base_executor
```
yifanmai, line 31: nit: two newlines after imports

tfx/components/base/base_executor_test.py
```diff
@@ -18,40 +18,153 @@
...
+    super(BeamExecutorTest, self).setUp()
+    self._input_data_path = (
+        os.path.join(os.environ.get(
+            'TEST_UNDECLARED_OUTPUTS_DIR', self.get_temp_dir()), 'input.txt'))
+    output_data_path = (
+        os.path.join(os.environ.get('TEST_UNDECLARED_OUTPUTS_DIR',
+                                    self.get_temp_dir()), 'output.txt'))
+    self._full_output_data_path = output_data_path + '-00000-of-00001'
+
+    self._generate_data(self._input_data_path)
```
yifanmai, line 129: nit: delete this file in tearDown()

tfx/components/base/base_executor_test.py
kshivvy: Added new line.

tfx/components/base/base_executor_test.py
kshivvy: I tried doing that, but I don't think there is a need to since the `super(BeamExecutorTest, self).tearDown()`method is automatically called and it removes `self._full_output_data_path` since it is a tmp file anyways. I got an error that said:

`Traceback (most recent call last):
  File "tfx/components/base/base_executor_test.py", line 144, in tearDown
    os.remove(self._full_output_data_path)
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/base_executor_testij6772g8/tmprgxyinl3/output.txt-00000-of-00001'`

PR #2285, opened 2020-08-06
Supporting Date spec in File Based ExampleGen driver

18jeffreyma:
Currently, in FileBasedExampleGen, users must specify span explicitly with the `{SPAN}` tag; however, users may have data organized by date stamp, with This CL adds support for parsing the calendar date (year, month, and day) of data with a set of Date specs `{YYYY}, {MM}, {DD}`. Using this calendar date, the span number is calculated as the number of days since unix epoch (`1970-01-01`).


tfx/components/example_gen/utils.py
```diff
@@ -346,12 +393,18 @@ def calculate_splits_fingerprint_span_and_version(
     logging.info('select span and version = (%s, %s)', select_span,
                  select_version)
     # Find most recent span and version for this split.
-    latest_span, latest_version = _retrieve_latest_span_version(
+    latest_span, latest_span_elems, latest_version, = _retrieve_latest_span_elems_version(
```
1025KB, line 396: my bad, i feel we should probably move split.pattern change in _retrieve_latest_span_elems_version, and just return span and version (to align, both span and version should be int, we might need to support {span:2} for the format (`01`, `02`), can be in future PR)

tfx/components/example_gen/utils.py
```diff
@@ -300,19 +332,34 @@ def _retrieve_latest_span_version(
             'Cannot find %s number from %s based on %s' %
             (VERSION_PROPERTY_NAME, file_path, split_regex_pattern))
 
-    if latest_span is None or span_int > int(latest_span):
+    if (latest_span_elems is None
+        or span_elems_int > [int(elem) for elem in latest_span_elems]):
```
1025KB, line 336: compare two list seems not ideal,  convert the date to span_int and still use span_int for comparison?

tfx/components/example_gen/utils.py
```diff
@@ -283,9 +315,9 @@ def _retrieve_latest_span_version(
     if result is None:
       raise ValueError('Glob pattern does not match regex pattern')
 
-    span_str = result.group(SPAN_PROPERTY_NAME)
+    span_elems_str = [result.group(name) for name in span_group_names]
```
1025KB, line 318: I feel have the span and date branch here might be more readable instead of using the magic `span_group_names ` and line348~360 

```
if is_match_span:
  span_strs = [...]
  span_int =
elif is_match_date:
  span_strs = 
  span_int = 

if span_int > xxx
```

tfx/components/example_gen/utils.py
18jeffreyma: should we just be setting them as int artifact properties?

tfx/components/example_gen/utils.py
18jeffreyma: fixed. now does conversion for each file path

tfx/components/example_gen/utils.py
18jeffreyma: refactored to have more clean branching logic.

tfx/components/example_gen/utils.py
18jeffreyma: updated PR to move split.pattern update inside `_retrieve_latest_span_version()`

18jeffreyma:
Resolved latest round of feedback thanks! Noteable changes:
- Made logic more clear by separating `is_match_span` and `is_match_date` from glob loop.
- Split pattern now updated from inside `_retrieve_latest_span_elems_version()`

casassg:
Could we add support for instead of only resolving with latest available date span, to resolve to the latest span available before Airflow execution_date (when using Airflow as scheduler). This would unlock using Airflow native backfill to generate datasets as they would have been available in X date.

Use case:
- I want to fo data validation on the dataset I trained my current outdated production model

18jeffreyma:
@casassg This PR is primarily for adding Date spec (and logic to parse Date spec) to pull latest data in FileBasedExampleGen. I believe support for processing specific (non-most recent) spans (and/or dates), will be added in a later PR.

casassg:
Sounds good. Looking forward to that one then. Will try to keep an eye

tfx/components/example_gen/utils.py
18jeffreyma: also updated return type to be ints, and driver to set span and version as int properties

18jeffreyma:
More updates:
- Span and Version are now returned as integers, and they are set as custom integer properties for Example artifacts.

RELEASE.md
```diff
@@ -44,6 +44,7 @@
 *   Added custom config to Transform component and made it available to
     pre-processing fn.
 *   Supported custom extractors in Evaluator.
+*   Supported `Date` spec in input config for file based ExampleGen.
```
1025KB, line 47: mention the type is changed in breaking change

18jeffreyma: updated, thanks!

18jeffreyma:
Resolved latest round of feedback, thanks!

tfx/components/example_gen/utils_test.py
```diff
@@ -427,8 +518,66 @@ def testCalculateSplitsFingerprintSpanAndVersion(self):
...
+    splits = [
+        example_gen_pb2.Input.Split(
+            name='s1', pattern='{YYYY}{MM}{DD}/ver{VERSION}/split1/*'),
+        example_gen_pb2.Input.Split(
+            name='s2', pattern='{YYYY}{MM}{DD}/ver{VERSION}/split2/*')
+    ]
+    _, span, version = utils.calculate_splits_fingerprint_span_and_version(
+        self._input_base_path, splits)
+    self.assertEqual(span, 2)
+    self.assertEqual(version, 2)
```
1025KB, line 580: is there a test that verify the split pattern updates of calculate_splits_fingerprint_span_and_version?

we want to make sure it's covered (especially for format like '02')

18jeffreyma: added tests to check for updated split patterns

tfx/components/example_gen/utils.py
```diff
@@ -225,13 +234,19 @@ def _retrieve_latest_span_version(
...
+  into a span number by counting the number of days since 01/01/1970.
+
   Args:
     uri: The base path from which files will be searched.
     split: An example_gen_pb2.Input.Split object which contains a split pattern,
       to be searched on.
 
   Returns:
-    Tuple of two strings, Span (optional) and Version (optional).
+    Tuple of two strings, Span (optional) and Version (optional). Note
```
hughmiao, line 247: nit, tuple of two int

tfx/components/example_gen/utils.py
```diff
@@ -241,27 +256,50 @@ def _retrieve_latest_span_version(
...
+    return (None, None)
 
-  if split.pattern.count(SPAN_SPEC) != 1:
+  if is_match_span and split.pattern.count(SPAN_SPEC) != 1:
     raise ValueError('Only one %s is allowed in %s' %
                      (SPAN_SPEC, split.pattern))
+  elif is_match_date and not all(split.pattern.count(spec) == 1
+                                 for spec in DATE_SPECS):
+    raise ValueError('Exactly one of each date spec is required in %s' %
+                     split.pattern)
```
hughmiao, line 282: nit, consider move all the validation to to some util

tfx/components/example_gen/driver.py
```diff
@@ -92,11 +92,11 @@ def _prepare_output_artifacts(
     example_artifact.set_string_custom_property(
         utils.FINGERPRINT_PROPERTY_NAME,
         exec_properties[utils.FINGERPRINT_PROPERTY_NAME])
-    example_artifact.set_string_custom_property(
+    example_artifact.set_int_custom_property(
```
hughmiao, line 95: note, this is breaking change, we shall still use string custom property for this?

/cc jyzhao@ any context of changing this to int?

18jeffreyma: @1025KB

tfx/components/example_gen/utils.py
18jeffreyma: fixed, thanks!

tfx/components/example_gen/utils.py
18jeffreyma: fixed. moved all initial spec finding validation into `_verify_split_specs()`, which returns bools indicating what should be matched

tfx/components/example_gen/driver.py
1025KB: Hui, we change to int because internal we use int?

or you suggest to keep the string custom property and later we add int property(not custom property) directly?

hughmiao: >  you suggest to keep the string custom property and later we add int property(not custom property) directly?

that looks a better approach. I think to unblock the cl, let's keep using string for now, and we can always do it in separate cls. wdyt?

18jeffreyma: sgtm. reverted back to using `set_string_custom_property()`. util functions like `calculate_splits_fingerprint_span_and_version()` still return ints, but `_prepare_output_artifacts()` in `driver.py` coverts int back to string when setting artifact property.

18jeffreyma:
resolved latest round of feedback, thanks! notable changes:
- split pattern spec preprocessing (and error raising) is now handled in a private function.
- span and version have been reverted back to string custom properties for example artifact

tfx/components/example_gen/utils.py
```diff
@@ -214,9 +223,39 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
+  is_match_version = VERSION_SPEC in split.pattern
+
+  if [is_match_span, is_match_date].count(True) > 1:
+    raise ValueError('Either span spec or date specs must be specified '
+                     'exclusively.')
+
+  if is_match_span and split.pattern.count(SPAN_SPEC) != 1:
+    raise ValueError('Only one %s is allowed in %s' % (SPAN_SPEC,
+                                                       split.pattern))
+  elif is_match_date and not all(split.pattern.count(spec) == 1
```
hughmiao, line 240: instead of `elif`, use a separate `if` ?

tfx/components/example_gen/utils.py
```diff
@@ -214,9 +223,39 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
+                                 for spec in DATE_SPECS):
+    raise ValueError('Exactly one of each date spec is required in %s' %
```
hughmiao, line 242: nit, add YYYY, MM, DD to the error string, which helps the users to fix their config

tfx/components/example_gen/utils.py
```diff
@@ -214,9 +223,39 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
+def _verify_split_pattern_specs(
+    split: example_gen_pb2.Input.Split) -> Tuple[bool, bool, bool]:
+  """Verify and identify specs to be matched in split pattern."""
+  is_match_span = SPAN_SPEC in split.pattern
+  is_match_date = any(spec in split.pattern for spec in DATE_SPECS)
+  is_match_version = VERSION_SPEC in split.pattern
+
+  if [is_match_span, is_match_date].count(True) > 1:
+    raise ValueError('Either span spec or date specs must be specified '
+                     'exclusively.')
```
hughmiao, line 235: add the split.pattern to the error string like other cases, which helps users to fix configs

tfx/components/example_gen/utils.py
```diff
@@ -214,9 +223,39 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
+    raise ValueError('Only one %s is allowed in %s' % (SPAN_SPEC,
+                                                       split.pattern))
+  elif is_match_date and not all(split.pattern.count(spec) == 1
+                                 for spec in DATE_SPECS):
+    raise ValueError('Exactly one of each date spec is required in %s' %
+                     split.pattern)
+
+  if is_match_version and (not is_match_span and not is_match_date):
+    raise ValueError('Version spec provided, but Span or Date spec is not '
+                      'present.')
```
hughmiao, line 247: ditto, + split.pattern

tfx/components/example_gen/utils.py
```diff
@@ -276,19 +322,40 @@ def _retrieve_latest_span_version(
...
+    if is_match_span:
+      span_strs = [result.group(SPAN_PROPERTY_NAME)]
+      try:
+        span_int = int(span_strs[0])
+      except ValueError:
+        raise ValueError('Cannot find %s number from %s based on %s' %
+                         (SPAN_PROPERTY_NAME, file_path, split_regex_pattern))
+
+    elif is_match_date:
+      span_strs = [result.group(name) for name in ['year', 'month', 'day']]
```
hughmiao, line 348: nit, for ['year', 'month', 'day'], using some constants?

tfx/components/example_gen/utils.py
```diff
@@ -276,19 +322,40 @@ def _retrieve_latest_span_version(
...
     if result is None:
       raise ValueError('Glob pattern does not match regex pattern')
 
-    span_str = result.group(SPAN_PROPERTY_NAME)
-    try:
-      span_int = int(span_str)
-    except ValueError:
-      raise ValueError('Cannot find %s number from %s based on %s' %
-                       (SPAN_PROPERTY_NAME, file_path, split_regex_pattern))
+    span_strs = None
```
hughmiao, line 336: nit, s/span_strs/matched_span_tokens

tfx/components/example_gen/utils.py
```diff
@@ -300,46 +367,60 @@ def _retrieve_latest_span_version(
...
+  elif is_match_date:
+    for spec, value in zip(DATE_SPECS, latest_span_elems):
+      split.pattern = split.pattern.replace(spec, value)
+
+  latest_version_int = None
+  if is_match_version:
+    split.pattern = split.pattern.replace(VERSION_SPEC, latest_version)
+    latest_version_int = int(latest_version)
+
+  return latest_span_int, latest_version_int
```
hughmiao, line 396: nit, the function still a bit long, probably decompose it to logical steps improve readability, 

e.g., 
_find_matched_span_tokens_from_path
_resolve_latestest_span_version

tfx/components/example_gen/utils.py
```diff
@@ -300,46 +367,60 @@ def _retrieve_latest_span_version(
...
+  {MM}, and {DD}, and optionally, the {VERSION} placeholder, attempts to find
+  aligned values that results in all splits having the most recent span and most
+  recent version for that span.
 
   Args:
     input_base_uri: The base path from which files will be searched.
-    splits: An iterable collection of example_gen_pb2.Input.Split objects. Note
-      that this function will update the {SPAN} in this and {VERSION} tags in
-      the split config to actual Span and Version numbers.
+    splits: An iterable collection of example_gen_pb2.Input.Split objects.
```
hughmiao, line 411: add the Note the tags will be updated part in the previous comment

RELEASE.md
```diff
@@ -45,6 +45,10 @@
     pre-processing fn.
 *   Supported custom extractors in Evaluator.
 *   Deprecated tensorflow dependency from MLMD python client.
+*   Supported `Date` spec in input config for file based ExampleGen.
+*   Span and Version are now stored as `int` custom properties for example
+    artifacts, where previously they were stored as strings.
```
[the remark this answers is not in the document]
18jeffreyma: apologies forgot to remove on revert, fixed.

tfx/components/example_gen/utils.py
18jeffreyma: fixed

tfx/components/example_gen/utils.py
18jeffreyma: updated

tfx/components/example_gen/utils.py
18jeffreyma: fixed

tfx/components/example_gen/utils.py
18jeffreyma: constraints are handled by the `datetime` package

tfx/components/example_gen/utils.py
18jeffreyma: updated.

tfx/components/example_gen/utils.py
18jeffreyma: updated, thanks!

tfx/components/example_gen/utils.py
18jeffreyma: updated. also changed `latest_span_elems` to `latest_span_tokens`

tfx/components/example_gen/utils.py
18jeffreyma: addressed. decomposed span matching as well as glob and regex generation to two different private functions.

18jeffreyma:
addressed above comments, thanks!

RELEASE.md
ruoyu90: I didn't find anything added under '## Breaking changes' section?

18jeffreyma: @ruoyu90 we decided to revert this type change. the proposed change was making span and version int custom properties, but since that would likely block the CL, we've decided to push it back.

https://github.com/tensorflow/tfx/pull/2285#discussion_r469564944

tfx/components/example_gen/utils.py
```diff
@@ -214,9 +223,123 @@ def _glob_to_regex(glob_pattern: Text) -> Text:
...
+    raise ValueError('Only one %s is allowed in %s' % (VERSION_SPEC,
+                                                       split.pattern))
+
+  return is_match_span, is_match_date, is_match_version
+
+
+def _find_matched_span_version_from_path(
+    file_path: Text, split_regex_pattern: Text, is_match_span: bool,
+    is_match_date: bool, is_match_version: bool
+) -> Tuple[Optional[List[Text]], Optional[int]]:
```
18jeffreyma, line 262: update typing

PR #2291, opened 2020-08-06
Modified ExampleGen and StatisticsGen executors to implement FuseableBeamExecutor class

kshivvy:
This PR is for `tfx/components/statistics_gen/executor.py` and `tfx/components/example_gen/base_example_gen_executor.py`.  Please review these files.

This PR depends on [the PR](https://github.com/tensorflow/tfx/pull/2274) for `tfx/components/base/base_executor.py` and `tfx/components/base/base_executor_test.py`.  Please reference the dependent PR as needed for changes in those files.

### DIFF
--- a/tfx/components/example_gen/base_example_gen_executor.py
+++ b/tfx/components/example_gen/base_example_gen_executor.py
@@ -22,7 +22,7 @@ import abc
 import bisect
 import hashlib
 import os
-from typing import Any, Dict, List, Text
+from typing import Any, Dict, List, Text, Union, Tuple
 
 import absl
 import apache_beam as beam
@@ -106,7 +106,7 @@ def _InputToSerializedExample(pipeline: beam.Pipeline,
 
 
 class BaseExampleGenExecutor(
-    with_metaclass(abc.ABCMeta, base_executor.BaseExecutor)):
+    with_metaclass(abc.ABCMeta, base_executor.FuseableBeamExecutor)):
   """Generic TFX example gen base executor.
 
   The base ExampleGen executor takes a configuration and converts external data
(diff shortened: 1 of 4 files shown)


kshivvy:
@charlesccychen 
@yifanmai

tfx/components/example_gen/base_example_gen_executor.py
```diff
@@ -131,6 +131,47 @@ class BaseExampleGenExecutor(
...
+                        ) -> Tuple[Dict[Text, type], Dict[Text, type]]:
+    return self.input_signature, self.output_signature
+
+  def read_inputs(self, pipeline: beam.Pipeline,
+                  input_dict: Dict[Text, List[types.Artifact]],
+                  output_dict: Dict[Text, List[types.Artifact]],
+                  exec_properties: Dict[Text, Union[int, float, Text]]
+                  ) -> Dict[Text, beam.pvalue.PCollection]:
+    inputs = self.GenerateExamplesByBeam(pipeline, input_dict, exec_properties)
+    return inputs
```
yifanmai, line 151: I don't think this is correct. I think ExampleGen should be modeled as a component that takes no inputs (because it is a "source" component). `GenerateExampleByBeam` should be placed in `run_component` instead.

tfx/components/example_gen/base_example_gen_executor.py
```diff
@@ -131,6 +131,47 @@ class BaseExampleGenExecutor(
...
+  def __init__(self):
+    super(BaseExampleGenExecutor, self).__init__()
+    self.input_signature = {}
+    self.output_signature = {}
+
+  def beam_io_signature(self, input_dict: Dict[Text, List[types.Artifact]],
+                        output_dict: Dict[Text, List[types.Artifact]],
+                        exec_properties: Dict[Text, Union[int, float, Text]]
+                        ) -> Tuple[Dict[Text, type], Dict[Text, type]]:
+    return self.input_signature, self.output_signature
```
yifanmai, line 143: You need to derive the signatures (i.e. the span names) from the `exec_properties` here. Right now, the signatures are only populated _after_ running the Beam pipelines. This produces a chicken and egg problem that will prevent you from doing fusion (you to know the signatures before running the pipeline to perform fusion, but the signatures are only available _after_ running the pipeline).

tfx/components/example_gen/base_example_gen_executor.py
```diff
@@ -131,6 +131,47 @@ class BaseExampleGenExecutor(
   ExampleGen can override `GenerateExamplesByBeam`.
   """
 
+  def __init__(self):
+    super(BaseExampleGenExecutor, self).__init__()
+    self.input_signature = {}
+    self.output_signature = {}
```
yifanmai, line 137: This shouldn't be required after changing how `beam_io_signature()` works; see my next comment.

tfx/components/example_gen/base_example_gen_executor.py
kshivvy: I placed `GenerateExampleByBeam` in `run_component` as suggested.

tfx/components/example_gen/base_example_gen_executor.py
kshivvy: Done.

tfx/components/example_gen/base_example_gen_executor.py
kshivvy: I derived the signatures from `exec_properties` inside `beam_io_signature()` as suggested- let me know if the new logic makes sense.

PR #2304, opened 2020-08-11
Adding support for width modifiers for Span and Version specs in File Based ExampleGen driver

18jeffreyma:
NOTE: This PR is branched from and dependent on PR#[2285](https://github.com/tensorflow/tfx/pull/2285). That PR will need to be approved and merged before this one can.

This CL adds support for explicit width modifiers for Span and Version specs. In the even that Span and Version specs do not have separators between them, with this PR, the user will be able to specify a width of characters in the Span/Version placeholder in the split config, which the regex parser will use to parse exactly that number of digits. Currently, the regex parser performs a greedy matching, which mean that, without separators between specs, a version digit may incorrectly be attributed to a span digit (and vice versa).

Example: `1234_test.data` with the pattern `{SPAN:2}{VERSION:2}_test.data` will have a span of 12 and a version number of 34 with new logic.


tfx/components/example_gen/utils.py
```diff
@@ -242,35 +261,108 @@ def _retrieve_latest_span_version(
...
+
+  elif is_match_date:
+    for spec in DATE_SPECS:
+      split_glob_pattern = split_glob_pattern.replace(spec, '*')
+    # Defines a clear number of digits for certain element of date. This covers
+    # cases where date stamps may not have seperators between them.
+    split_regex_pattern = split_regex_pattern.replace(
+        YEAR_SPEC, '(?P<{}>.*)'.format('year'))
+    split_regex_pattern = split_regex_pattern.replace(
+        MONTH_SPEC, '(?P<{}>.{{2}})'.format('month'))
```
1025KB, line 329: is this {2} for width? how about years?

tfx/components/example_gen/utils.py
```diff
@@ -242,35 +261,108 @@ def _retrieve_latest_span_version(
...
+    span_width_regex = '.*'
+    if len(span_arg_match[0]) > 0:
+      # Check for proper formatting of span spec with width modifier.
+      # This spec should have the form '{SPAN:xxx}', where the span regex
+      # captures the sequence ':xxx'.
+      if span_arg_match[0][0] != ':':
+        raise ValueError('Span spec with width modifier is improperly '
+                         'formatted: %s' % split.pattern)
+      try:
+        span_width = int(span_arg_match[0][1:])
```
1025KB, line 308: can you try get this just like how you get specs?

e.g. use {SPAN:?<width>} instead of {SPAN(.*?)}?

tfx/components/example_gen/utils.py
18jeffreyma: i can add that back in the date PR

tfx/components/example_gen/utils.py
18jeffreyma: I can try it that way, sure. The main reason I did it this way is so we can just have one matching for both `{SPAN}` and `{SPAN:x}`. If we have the capture group to capture the integer, it adds some case work to the code, which i'm wasn't sure about. For example, to check is_match_span, we need to check if `{SPAN}` is in split pattern or if `{SPAN:(.*)}` regex matches the split pattern. any thoughts on this?

18jeffreyma: updated this. new behavior now uses an or regex. for example, for span the regex is now `{SPAN}|{SPAN:(.+?)}` which matches either exactly `{SPAN}` or `{SPAN:x}` (where `x` is a 1+ length string). also removed the tests for testing malformed specs like `{SPANx}` since they don't seem like relevant cases we need to account for.

18jeffreyma:
addressed most recent round of comments thanks!

tfx/components/example_gen/utils.py
1025KB: you just need to get the width count from a string, right?
can we use the way like how you get specs (get yyyy number from string)?

18jeffreyma: fixed.  new logic is to used a named group, optimistically search for this named group, and use default greedy regex when named group does not exist

tfx/proto/example_gen.proto
```diff
@@ -46,6 +46,11 @@ message Input {
   //   - This spec matches digits and maps the data into the relevant SPAN
   //     numbers. e.g. 'data_{SPAN}-*.tfrecord' will collect files like
   //     'data_12-a.tfrecord', 'date_12-b.tfrecord'.
+  //   - Optionally, it could be specified with the width of the integers
+  //     when mapped.
+  //     e.g. 'data_{SPAN:2}.file' mas to files like 'data_02.file' and
```
hughmiao, line 51: nit, s/`mas`/`maps`

tfx/proto/example_gen.proto
```diff
@@ -46,6 +46,11 @@ message Input {
...
+  //       'data_27.file' (as inputs for Apan-2 and Span-2 respectively),
```
hughmiao, line 52: nit, s/Apan-2/Span-2 and s/Span-2/Span-27

tfx/proto/example_gen.proto
```diff
@@ -71,10 +76,10 @@ message Input {
   //     latest version for the latest span, and store the version number in
   //     metadata.
   //   - If VERSION is specified, but not SPAN, an error will be thrown.
+  //   - It could also be optionally specified with the width in the same way
```
hughmiao, line 79: under line 73, +
"//     e.g. 'data_{SPAN}/v{VERSION}/*' matches to files under 'data_0/v1/'
 //       to consume as input for Version-1 of Span-0."

tfx/proto/example_gen.proto
```diff
@@ -71,10 +76,10 @@ message Input {
...
+  //     as SPAN spec. e.g. 'data_span_{SPAN}_version_{VERSION:4}.data-*'.
```
hughmiao, line 80: after 80 
+ For zero padding digits (e.g., `0012`), the width must be used.

tfx/proto/example_gen.proto
18jeffreyma: fixed, apologies for typos

tfx/proto/example_gen.proto
18jeffreyma: fixed

tfx/proto/example_gen.proto
18jeffreyma: @hughmiao is this necessarily true though? I thought the OSS build of TFX has the ability to handle zero padding without specifying width.

18jeffreyma:
addressed most recent round of feedback (with just one optional topic for discussion), thanks!

tfx/proto/example_gen.proto
hughmiao: this is about the conversation we had last time, when in some mode, e.g., trigger example-gen with static range, we need to have the width information in order to reconstruct the path.

18jeffreyma: gotcha i see, this is a precaution for that behavior, thanks!

tfx/components/example_gen/utils.py
```diff
@@ -51,10 +51,20 @@
 SPAN_PROPERTY_NAME = 'span'
 # Span spec used in split pattern.
 SPAN_SPEC = '{SPAN}'
+# Span spec regex to capture width modifier. This matches the spec '{SPAN:x}'
+# and captures the string 'x'.
+SPAN_SPEC_WIDTH_REGEX = '{SPAN:(?P<width>.*?)}'
+# Full regex for matching span specs with or without width modifier.
+SPAN_FULL_REGEX = "{}|{}".format(SPAN_SPEC, SPAN_SPEC_WIDTH_REGEX)
```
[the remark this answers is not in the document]
ruoyu90: Actually +1 on separating them to increase readability :)

1025KB:
let's not commit after I approved, otherwise there will be some issue of merging the PR

PR #2317, opened 2020-08-12
Update Metadata gRPC env in default kubeflow metadata config

Jeffwan:
Kubernetes by default inject service host and port as environment variables into pod container. 

```
$ kubectl get svc                                      
NAME                                                TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)                               AGE
metadata-envoy-service                              ClusterIP   10.100.80.21     <none>        9090/TCP                              4d18h
metadata-grpc-service                               ClusterIP   10.100.30.11     <none>        8080/TCP                              4d18h
```

The naming convention is  
-`SERVICE_NAME` + `_SERVICE_HOST` 
-`SERVICE_NAME` + `_SERVICE_PORT`

In this case, it should be `METADATA_GRPC_SERVICE_SERVICE_HOST ` and `METADATA_GRPC_SERVICE_SERVICE_PORT `

User can override this for sure. I think it's better to make it consistent as kubeflow setup. `get_default_kubeflow_metadata_config ()` will help populate the right environment.

Upstream manifest:
- Kubeflow https://github.com/kubeflow/manifests/blob/master/pipeline/upstream/base/metadata/metadata-grpc-service.yaml#L6
- Kubeflow Pipeline Standalone https://github.com/kubeflow/pipelines/blob/514120167e2e255d6c35cb564fd57ee75b452f56/manifests/kustomize/base/metadata/metadata-grpc-service.yaml#L6

### DIFF
--- a/tfx/orchestration/kubeflow/container_entrypoint_test.py
+++ b/tfx/orchestration/kubeflow/container_entrypoint_test.py
@@ -61,13 +61,13 @@ class MLMDConfigTest(tf.test.TestCase):
 
   def testGrpcMetadataConnectionConfig(self):
     self._set_required_env_vars({
-        'METADATA_GRPC_SERVICE_HOST': 'metadata-grpc',
-        'METADATA_GRPC_SERVICE_PORT': '8080',
+        'METADATA_GRPC_SERVICE_SERVICE_HOST': 'metadata-grpc-service',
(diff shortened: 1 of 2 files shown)


Jeffwan:
Seems the error is unrelated. Is there's a way to retest it? 

/retest

PR #2296, opened 2020-08-07
Adding beam pipeline for distributed inference

JasonZhu01:
OSS Intern Project: Map distributed inference graph to beam pipeline.

Shortly speaking, we want to demonstrate that models with remote graphs can be partitioned, arranged, and executed in a beam pipeline (for distributed inference). The results from the beam pipeline should be the same as the original TF model's results.

The last PR (https://github.com/tensorflow/tfx/pull/2157) introduces the graph partitioning part. The beam pipeline in this PR arranges and executes the partitioned subgraphs.

Note that this PR and the last PR both operate on GraphDefs instead of SavedModels for simplicity. The next PR will operate on SavedModels.

PR #2157, opened 2020-07-15
Adding subgraph partitioning strategy for distributed inference

JasonZhu01:
OSS Intern Project: Map distributed inference graph to beam pipeline.

Shortly speaking, we want to demonstrate that models with remote graphs can be partitioned and arranged in a beam pipeline. The results from the beam pipeline should be the same as the original TF model's results.

First working with GraphDefs. Tried two approaches: 1. op-level partitioning, 2. subgraph partitioning.
We organized each approach into two parts: the graph partitioning part and the beam pipeline part.

This PR focuses on the graph partitioning part of the subgraph partitioning strategy.

PR #2324, opened 2020-08-13
Adding SavedModel experiments for distributed inference

JasonZhu01:
OSS Intern Project: Map distributed inference graph to beam pipeline.

Shortly speaking, we want to demonstrate that models with remote graphs can be partitioned and arranged in a beam pipeline. The results from the beam pipeline should be the same as the original TF model's results.

After working with GraphDefs (PRs: https://github.com/tensorflow/tfx/pull/2157, https://github.com/tensorflow/tfx/pull/2296), we moved our attention to SavedModels.

In this PR, the create_complex_graph file saves the example graphs as SavedModels. The graph partitioning library takes in SavedModel directory paths, extracts `GraphDef` protos and output names for each graph, invokes partition_all() from graphdef_experiments/subgraph_partitioning, then saves the partitioned subgraphs into SavedModels. The beam pipeline library takes in partitioned SavedModel directory paths and builds a PTransform that executes the partitioned subgraphs. The results from the beam pipeline and the results from the original model are indeed the same.

After experimenting with GraphDefs and SavedModels, I realized that a much better/modular code structure can be designed. Hopefully in the future, someone (maybe me) will use these experiments as a starting point of building a feature for production.

### DIFF
--- a/tfx/experimental/distributed_inference/savedmodel_experiments/graph_partition.py
+++ b/tfx/experimental/distributed_inference/savedmodel_experiments/graph_partition.py
@@ -11,206 +11,501 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Graph partitioning on TensorFlow SavedModels with remote ops.
+"""Graph partitioning on TensorFlow GraphDefs with remote ops.
 
-The current implementation invokes the subgraph partitioning algorithm.
+The current partitioning strategy aims for maximum subgraphs.
 
 There are two libraries representing two stages: graph_partition and
-beam_pipeline. In this library, we take in some SavedModel inputs, partition
-the graphs, and store the partitioned subgraphs into lists of SavedModel
-directory paths. The order of the list represents the order of execution.
-Take a look at beam_pipeline if you want to run the partitioned subgraphs.
+beam_pipeline. In this library, we take in some GraphDef inputs, partition
+the graphs, and store the partitioned subgraphs into lists of ExecutionSpecs.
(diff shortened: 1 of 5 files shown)


JasonZhu01:
@goutham @yifanmai

tfx/experimental/distributed_inference/savedmodel_experiments/graph_partition.py
```diff
@@ -11,501 +11,206 @@
...
+
 from tfx.experimental.distributed_inference.graphdef_experiments.subgraph_partitioning import execution_spec
+from tfx.experimental.distributed_inference.graphdef_experiments.subgraph_partitioning import graph_partition
 
 
-def get_graph_name_to_graph_def(
-    graph_name_to_filepath: Mapping[Text, Text]
-) -> Dict[Text, tf.compat.v1.GraphDef]:
-  """Gets the `GraphDef` protos from files.
+def get_info_from_paths(
```
goutham, line 54: This has two functions in 1. Don't see a good reason to return a tuple.

tfx/experimental/distributed_inference/savedmodel_experiments/graph_partition_test.py
```diff
@@ -17,111 +17,47 @@
...
 
+  def test_with_golden_set(self):
+    """Tests if the SavedModels match the golden set.
 
-def _get_node_names(graph_def):
-  return {node.name for node in graph_def.node}
+    Problem: Saved_model.save or saved_model.simple_save doesn't allow us
+             to specify the format (binary or text) of the saved_model file
+             inside the folder. Not sure if I should use binary files
+             for the golden set.
```
goutham, line 57: Once saved with saved_model.save, it can be loaded as a proto and written back in text format for the golden set.

PR #2339, opened 2020-08-15
Adding RangeConfig proto and support in file based ExampleGen

18jeffreyma:
NOTE: This PR is dependent on [PR#2304](https://github.com/tensorflow/tfx/pull/2304). That PR must be merged before this PR can be merged.

Currently, file based ExampleGen default behavior is to use `{SPAN}` spec to retrieve the latest span possible and ingest it. This CL adds a new abstract RangeConfig proto as well as support for an optional `RangeConfig` parameter for file based ExampleGen to specify a specific static range of span values to search through or a set of spans to exclude from consideration. 

Additionally, added a public `date_to_span_number()` method so users can easily convert calendar dates to span numbers (will make it easier for users to construct their own range_configs as well).

This CL will both allow users to process specific spans of data, but will also allow them to do so more efficiently. In the event, users have a very large collection of files, previously, they would need to glob through all of the possible files, whereas now, a user can specify a specific span, which TFX will replace the spec with, eliminating the need to filter through all files.


18jeffreyma:
@1025KB

tfx/proto/range_config.proto
```diff
@@ -0,0 +1,52 @@
...
+package tfx.configs;
+
+// Describes a static window within the specified span numbers
+// [start_span_number, end_span_number]. If not all spans in the
+// range are available, processing may continue on a partial range as long as 
+// RangeConfig.min_spans are availiable.
+// Note that both numbers should be specified for StaticRange.
+message StaticRange {
+  int32 start_span_number = 1;
+  int32 end_span_number = 2;
```
1025KB, line 25: note that currently we only support start_span_number == end_span_number as examplegen only process one span at a time

tfx/components/example_gen/utils.py
```diff
@@ -397,6 +409,17 @@ def _retrieve_latest_span_version(
                                              is_match_span, is_match_date,
                                              is_match_version))
 
+    # Check parsed span number against range config, if it exists.
+    if range_config:
+      if range_config.HasField('static_range'):
+        if (match_span_int < range_config.static_range.start_span_number or
+            match_span_int > range_config.static_range.end_span_number):
+          # If span not in static range, skip to next file path.
+          continue
```
1025KB, line 418: for this, we don't want to scan the files, we just want to fill the span (with width) into the span spec or date spec, and directly access that span, if not available, just error

tfx/proto/range_config.proto
18jeffreyma: gotcha, i will update the behavior

tfx/components/example_gen/utils.py
18jeffreyma: @hughmiao a couple edge cases to consider. if range_config specifies span number of 1 and split pattern is `span{SPAN}/*`, if we did a naive substitution (just replace spec with int), this would fail for a filepath like `span0001/data`. May need some discussion on whether this behavior is intended or not.

18jeffreyma:
added range_config to all file based example gen sub classes. regarding 2nd comment, some more discussion is needed to move forward.

18jeffreyma:
since range config will perform substitution using width, this PR has now been made dependent to Width PR.

tfx/components/example_gen/utils.py
18jeffreyma: additionally, since we are doing replacement and not filtering, should we also add a RangeConfig for date logic? it seems like our range_config logic in substituting won't allow users to process specific dates?

18jeffreyma:
updated PR to correct behavior. `RangeConfig.static_range` (if it exists) will replace the span spec in the glob and regex patterns (using the width modifier, if it exists, to correctly pad the number of zeros needed). Thus, we no longer sweep through the entire files, and we specifically glob the files that are relevant based on this range_config. PTAL.

tfx/proto/example_gen.proto
```diff
@@ -56,25 +62,28 @@ message Input {
...
   //   - If VERSION is specified, but not SPAN, an error will be thrown.
-  //
+  //   - It could also be optionally specified with the width in the same way
+  //     as SPAN spec. e.g. 'data_span_{SPAN}_version_{VERSION:4}.data-*'.
+  //     For zero padding digits (e.g., `0012`), the width must be used.
+  // 
   // TODO(b/162622803): add default behavior for when version spec not present.
-  // TODO(jyzhao): support retrieving a specific number of digits in Span and
-  //     Version spec (i.e. {SPAN:2} or {VERSION:4})
   // TODO(jyzhao): support range config for define certain spans to process.
```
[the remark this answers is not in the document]
18jeffreyma: PR now synced to head (with width PR merged), so this diff is gone.

tfx/components/example_gen/utils.py
```diff
@@ -272,6 +288,17 @@ def _verify_split_pattern_specs(
...
+    if range_config.HasField('static_range'):
+      # For ExampleGen, RangeConfig must specify an exact span to look for,
+      # since only one span is processed at a time.
+      start_span_number = range_config.static_range.start_span_number
+      end_span_number = range_config.static_range.end_span_number
+      if start_span_number != end_span_number:
+        raise ValueError(
+            'Start and end span numbers for RangeConfig.static_range must '
+            'be equal: (%s, %s)' % (start_span_number, end_span_number))
+
```
hughmiao, line 301: separate this to another validation function? this looks loose coupled with rest of the code in this util

tfx/components/example_gen/utils.py
```diff
@@ -272,6 +288,17 @@ def _verify_split_pattern_specs(
     raise ValueError('Only one %s is allowed in %s' %
                      (VERSION_SPEC, split.pattern))
 
+  if range_config:
+    if range_config.HasField('static_range'):
```
hughmiao, line 292: nit, range_config and range_config.HasField('static_range') ?

tfx/components/example_gen/utils.py
```diff
@@ -335,37 +363,68 @@ def _create_matching_glob_and_regex(
...
+      if span_width_str:
+        span_width_int = int(span_width_str)
+        if span_width_int < len(span_str):
+          raise ValueError(
+              'Span spec width is less than number of digits in span: (%s, %s)'
+              % (span_width_int, span_str))
+        span_str = span_str.zfill(span_width_int)
+
+      span_regex_replace = span_str
+      span_glob_replace = span_str
```
hughmiao, line 395: consider to move line 382 - 395, line 411 - 419 to a util, e.g., _replace_split_glob_and_regex_, which helps the reader to grasp the overview 

in addition, this util can be used in the callsite, e.g., _create_matching_glob_and_regex, then 
```
_create_matching_glob_and_regex
if range_config:
   _replace_split_glob_and_regex(range-config)
```

tfx/components/example_gen/utils.py
```diff
@@ -395,7 +454,9 @@ def _create_matching_glob_and_regex(
 
 def _retrieve_latest_span_version(
     uri: Text,
-    split: example_gen_pb2.Input.Split) -> Tuple[Optional[int], Optional[int]]:
+    split: example_gen_pb2.Input.Split,
+    range_config: Optional[range_config_pb2.RangeConfig] = None
```
hughmiao, line 458: change the name to _get_target_span_version?
as it may not look for latest span & version, if the range-config is present

tfx/components/example_gen/utils.py
```diff
@@ -516,7 +584,7 @@ def calculate_splits_fingerprint_span_and_version(
                  select_version)
     # Find most recent span and version for this split.
     latest_span, latest_version = _retrieve_latest_span_version(
-        input_base_uri, split)
+        input_base_uri, split, range_config=range_config)
```
hughmiao, line 587: ditto, change the latest_span, latest_version to be target_span, target_version

hughmiao (changes requested):
thanks, Jeffrey. Overall lgtm, comments to reorganize to the code a bit.

tfx/components/example_gen/utils.py
18jeffreyma: the main issue i see with using this in the call site is that when we create the glob pattern for a pattern say `span{SPAN}/ver{VERSION}/*` the resulting glob is `span*/ver*/*`. There doesn't seem to be a clean way to identify which asterisk was originally the span tag and which was the version tag, which would make subsequent substitution in the glob a bit messy. Was there a different logic you had in mind for this callsite usage?

18jeffreyma: refactored out width into a seperate method as well as range_config replacement logic

tfx/components/example_gen/driver_test.py
```diff
@@ -127,6 +129,46 @@ def testResolveExecProperties(self):
...
+    )
+    updated_input_config = example_gen_pb2.Input()
+    json_format.Parse(self._exec_properties[utils.INPUT_CONFIG_KEY],
+                      updated_input_config)
+    # Check if correct span inside static range is selected.
+    self.assertProtoEquals(
+        """
+        splits {
+          name: "s1"
+          pattern: "span01/version01/split1/*"
```
hughmiao, line 165: qq, in this case the pattern: span{SPAN:2}/version{VERSION} the version does not have a width config. should the expected string : span01/version1/split1/

is the zero padding some default behavior? or we carry it over from SPAN:2 ?

if it is the former, consider to update docstring / document, if the the latter, this is an incorrect behavior, consider to address these issues in a separate CL.

tfx/components/example_gen/utils.py
```diff
@@ -393,9 +476,11 @@ def _create_matching_glob_and_regex(
   return split_glob_pattern, split_regex_pattern
 
 
-def _retrieve_latest_span_version(
+def _get_target_span_version(
     uri: Text,
-    split: example_gen_pb2.Input.Split) -> Tuple[Optional[int], Optional[int]]:
+    split: example_gen_pb2.Input.Split,
+    range_config: Optional[range_config_pb2.RangeConfig] = None
+) -> Tuple[Optional[int], Optional[int]]:
```
hughmiao, line 483: Please update the docstring of this function too, it is not always `recent` span / version anymore.

tfx/components/example_gen/utils.py
```diff
@@ -324,9 +353,60 @@ def _find_matched_span_version_from_path(
...
+  else:
+    raise ValueError("Only static_range in RangeConfig is supported.")
+
+
 def _create_matching_glob_and_regex(
     uri: Text, split: example_gen_pb2.Input.Split, is_match_span: bool,
-    is_match_date: bool, is_match_version: bool) -> Tuple[Text, Text]:
+    is_match_date: bool, is_match_version: bool,
+    range_config: Optional[range_config_pb2.RangeConfig]
+) -> Tuple[Text, Text]:
```
hughmiao, line 409: given this function is quite complex, consider to elaborate the docstring for future readers.

hughmiao (approved):
thanks, Jeffrey. overall lgtm. left some nit comments to update the docstring, and some non-blocking comment to address the width for VERSION.

tfx/components/example_gen/driver_test.py
18jeffreyma: it does not carry over from `{SPAN:2}`.

look at previous usage of the `{SPAN}` spec, before i introduced width modifier (and the version spec), span was retrieved by doing a greedy regex capture (and retrieving the padding as well). this is different from internal behavior i think, since based on proto docstring, internal requires width with any kind of zero padding, while OSS will just try to capture the entire zero padded span number for you.

IIRC, by default, TFX OSS tries to capture the entire span number with zero padding using the greedy regex, so maybe we need to update docstring.

18jeffreyma:
addressed latest feedback, need some quick discussion on docstring for example_gen. internal requires width info to zero pad, while OSS does not have this restriction (since width was introduced later).

18jeffreyma:
since TFX tries to grab zero padding digits by default, current documentation is not correct. updated proto docstring to be more accurate.

tfx/proto/range_config.proto
```diff
@@ -0,0 +1,37 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+syntax = "proto3";
+
+package tfx.configs;
+
+// Describes a static window within the specified span numbers
+// [start_span_number, end_span_number]. If not all spans in the
+// range are available, processing may continue on a partial range as long as 
+// RangeConfig.min_spans are availiable.
```
ruoyu90, line 21: `min_spans` is not available in the definition.

tfx/components/example_gen/utils.py
```diff
@@ -275,6 +291,20 @@ def _verify_split_pattern_specs(
   return is_match_span, is_match_date, is_match_version
 
 
+def _verify_range_config(
```
ruoyu90, line 294: Let's call out the restriction on `RangeConfig` in the component interface. I was thinking that we get the latest span that is inside the range specified by the `RangeConfig` until see the code here.

@1025KB  @hughmiao any context we go with the decision to allow only `start_span_number == end_span_number`? Another question will be, how do we support other range config types in ExampleGen?

ruoyu90 (commented):
Thanks Jeffrey!

tfx/proto/example_gen.proto
```diff
@@ -81,7 +80,6 @@ message Input {
   //   - If VERSION is specified, but not SPAN, an error will be thrown.
   //   - It could also be optionally specified with the width in the same way
   //     as SPAN spec. e.g. 'data_span_{SPAN}_version_{VERSION:4}.data-*'.
-  //     For zero padding digits (e.g., `0012`), the width must be used.
```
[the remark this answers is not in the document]
18jeffreyma: this is from discussion with Hui, above (we can also cover it at our sync). TFX does not require width for zero-padding retrieval, so unsure as to how to express this in docstring?

tfx/proto/range_config.proto
18jeffreyma: will update this docstring

tfx/proto/example_gen.proto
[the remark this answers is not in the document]
18jeffreyma: will discuss at tmr sync before i update this again.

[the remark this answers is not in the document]
18jeffreyma: synced with hui, will just keep this docstring as is

tfx/components/example_gen/utils.py
18jeffreyma: will move this check to the component interface. the 2nd portion was partially resolved at our sync, but i will leave open in case more discussion is needed.

18jeffreyma:
resolved latest round of feedback thanks! moved range_config.static_range check to component constructor (for better visibility) as well as reverted docstring back to original.

1025KB (approved):

build/BUILD
```diff
@@ -30,6 +30,7 @@ sh_binary(
         "//tfx/proto:trainer_pb2.py",
         "//tfx/proto:transform_pb2.py",
         "//tfx/proto:tuner_pb2.py",
+        "//tfx/proto:range_config_pb2.py",
```
1025KB, line 33: in alphabet order, move this to line 30

1025KB (commented):
could you sync to head for this PR, thanks!

build/BUILD
18jeffreyma: fixed

18jeffreyma:
synced to head, thanks!

1025KB (approved):

tfx/components/example_gen/utils_test.py
```diff
@@ -664,6 +666,100 @@ def testCalculateSplitsFingerprintSpanAndVersionWithDate(self):
...
+
+    range_config = range_config_pb2.RangeConfig(
+        static_range=range_config_pb2.StaticRange(start_span_number=2,
+                                                  end_span_number=2))
+    splits = [
+        example_gen_pb2.Input.Split(
+            name='s1', pattern='span{SPAN:2}/split1/*')
+    ]
+
+    with self.assertRaises(tf.errors.NotFoundError):
```
1025KB, line 683: why it's notFoundError instead of ValueError

18jeffreyma: b/c the glob pattern generated wont exist on file system, so iirc tf.gfile.glob will return a notFoundError.

PR #2397, opened 2020-08-27
Supporting multiple artifact input and output for Transform component

18jeffreyma:
Currently, Transform can only process input channels with single artifacts and output channels with single artifacts. This CL adds support for input and output channels with multiple artifacts to support transforming multiple example artifacts with one component run, making use of the previously introduced `get_split_uris()` util function.


18jeffreyma:
@1025KB

tfx/components/transform/executor.py
```diff
@@ -331,8 +331,7 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
       splits_config.analyze.append('train')
       splits_config.transform.extend(
           artifact_utils.decode_split_names(
-              artifact_utils.get_single_instance(
-                  input_dict[EXAMPLES_KEY]).split_names))
+              input_dict[EXAMPLES_KEY][0].split_names))
```
1025KB, line 334: let's use the union set of all splits for transform (or verify all input has the same splits name)?

tfx/components/transform/executor.py
```diff
@@ -350,27 +349,38 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+              'Transform split does not exist over all example artifacts: %s' %
+              split)
+        for data_uri in data_uris:
+          transform_data_paths.append(io_utils.all_files_pattern(data_uri))
+
+        transformed_example_uris = artifact_utils.get_split_uris(
+            output_dict[TRANSFORMED_EXAMPLES_KEY], split)
+        for output_uri in transformed_example_uris:
+          materialize_output_paths.append(
+              os.path.join(output_uri, _DEFAULT_TRANSFORMED_EXAMPLES_PREFIX))
```
1025KB, line 383: just confirm, can this makes sure one input artifact maps to one output artifact?

e.g.,

artifact1:train_split -> output artifact1:train_split
artifact1:eval_split -> output artifact1:eval_split
artifact2:train_split -> output artifact2: train_split
artifact2:eval_split -> output artifact2: eval_split

it would be good to have a test that input split has different row count, and verify output split row count

tfx/components/transform/executor.py
18jeffreyma: i think the latter option will probably be more well defined (and be more clear to users in case of an err), will fix that

tfx/components/transform/executor.py
18jeffreyma: i was thinking of just loading in the non-transformed files and comparing the count to the transformed files. is that what you're suggsting, or maybe something else?

18jeffreyma: updated tests. new behavior is using a temp directory for testdata and creating two artifact directories (one with double the data of the other to more easily verify 1-to-1 mapping). in `verify_transform_output` for each corresponding pair of input and output artifacts, we verify that the number of records in a certain split of the input artifact is equal to the number of records in the same split of the output transformed artifact.

18jeffreyma: TODO(jjma): test Transform API if uri count is not matched for input and output

tfx/components/transform/executor.py
```diff
@@ -350,27 +360,38 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
     if output_dict.get(TRANSFORMED_EXAMPLES_KEY) is not None:
-      transformed_example_artifact = artifact_utils.get_single_instance(
-          output_dict[TRANSFORMED_EXAMPLES_KEY])
-      # TODO(b/161490287): move the split_names setting to executor for all
-      # components.
-      transformed_example_artifact.split_names = (
-          artifact_utils.encode_split_names(list(splits_config.transform)))
+      for transformed_example_artifact in output_dict[TRANSFORMED_EXAMPLES_KEY]:
+        # TODO(b/161490287): move the split_names setting to executor for all
+        # components.
```
1025KB, line 376: we can remove this comment

tfx/components/transform/executor.py
```diff
@@ -350,27 +360,38 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
-            artifact_utils.get_split_uri(output_dict[TRANSFORMED_EXAMPLES_KEY],
-                                         split),
-            _DEFAULT_TRANSFORMED_EXAMPLES_PREFIX)
-        materialize_output_paths.append(transformed_example)
+        data_uris = artifact_utils.get_split_uris(input_dict[EXAMPLES_KEY],
+                                                  split)
+        if len(data_uris) != len(input_dict[EXAMPLES_KEY]):
+          raise ValueError(
+              'Transform split does not exist over all example artifacts: %s' %
+              split)
```
1025KB, line 386: maybe this should be in the get_split_uris function?

tfx/components/transform/executor_test.py
```diff
@@ -41,10 +43,39 @@ class _TempPath(types.Artifact):
 # TODO(b/122478841): Add more detailed tests.
 class ExecutorTest(tft_unit.TransformTestCase):
 
-  def _get_source_data_dir(self):
-    return os.path.join(
+  @classmethod
```
1025KB, line 46: I did some updates to this file, could you sync to head?

tfx/components/transform/executor_test.py
```diff
@@ -41,10 +43,39 @@ class _TempPath(types.Artifact):
...
+  def setUpClass(cls):
+    cls._temp_data_dir = tempfile.mkdtemp()
```
1025KB, line 48: use [this](https://github.com/tensorflow/tfx/blob/master/tfx/components/example_gen/import_example_gen/executor_test.py#L81) to get a temp output dir

tfx/components/transform/executor_test.py
```diff
@@ -41,10 +43,39 @@ class _TempPath(types.Artifact):
...
         os.path.dirname(os.path.dirname(__file__)), 'testdata')
 
+    source_example_dir = os.path.join(source_data_dir, 'csv_example_gen')
+    source_schema_dir = os.path.join(source_data_dir, 'schema_gen')
+    source_module_dir = os.path.join(source_data_dir, 'module_file')
+
+    io_utils.copy_dir(source_schema_dir,
+                      os.path.join(cls._temp_data_dir, 'schema_gen'))
+    io_utils.copy_dir(source_module_dir,
+                      os.path.join(cls._temp_data_dir, 'module_file'))
```
1025KB, line 60: is that necessary to copy schema and model_file?

we just need to create a new input examplegen artifact2 that point to the copy location, other file path can remain the same?

tfx/components/transform/executor.py
18jeffreyma: updated util function to do the length check.

tfx/components/transform/executor_test.py
18jeffreyma: yup, fixed this. doing some additional function cleaning

tfx/components/transform/executor_test.py
18jeffreyma: since setUpClass is a class method, we cant call self.get_temp_dir() to get a temp directory (so we use the tempfile method. Will add a tearDownClass for cleanup.

tfx/components/transform/executor_test.py
18jeffreyma: synced. now using `_get_dataset_size()` method.

18jeffreyma:
addressed latest round of feedback, thanks!

tfx/components/transform/executor_test.py
```diff
@@ -58,9 +60,29 @@ class _TempPath(types.Artifact):
...
+    # second artifact has twice as many as first.
+    artifact2_train_dir = os.path.join(cls._artifact2_uri, 'train')
+    for filename in tf.io.gfile.listdir(artifact2_train_dir):
+      io_utils.copy_file(os.path.join(artifact2_train_dir, filename),
+                         os.path.join(artifact2_train_dir, 'dup_' + filename))
+
+    artifact2_eval_dir = os.path.join(cls._artifact2_uri, 'eval')
+    for filename in tf.io.gfile.listdir(artifact2_eval_dir):
+      io_utils.copy_file(os.path.join(artifact2_eval_dir, filename),
+                         os.path.join(artifact2_eval_dir, 'dup_' + filename))
```
1025KB, line 85: you can just glob all files under cls._artifact2_uri and use one for loop to do duplication I think

and since train:eval = 2:1, the artifact1:artifact2= 2:1, will artifact1train = artifact2eval, or it's slightly different

18jeffreyma: i think its slightly different, the numbers are below

artifact1 train 9909
artifact1 eval 5091
artifact2 train 19818
artifact2 eval 10182

18jeffreyma: updated the behavior to use a single glob

tfx/components/transform/executor.py
18jeffreyma: https://github.com/tensorflow/tfx/blob/ff6917997340401570d05a4d3ebd6e8ab5760495/tfx/components/transform/executor.py#L904
https://github.com/tensorflow/tfx/blob/ff6917997340401570d05a4d3ebd6e8ab5760495/tfx/components/transform/executor.py#L1323
re: our discussion during sync, `_MakeDatasetList()` checks that the number of patterns matches output paths

tfx/components/transform/executor.py
```diff
@@ -329,10 +329,20 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
         raise ValueError('analyze cannot be empty when splits_config is set.')
     else:
       splits_config.analyze.append('train')
-      splits_config.transform.extend(
-          artifact_utils.decode_split_names(
-              artifact_utils.get_single_instance(
-                  input_dict[EXAMPLES_KEY]).split_names))
+      split_names = artifact_utils.decode_split_names(
+          input_dict[EXAMPLES_KEY][0].split_names)
```
zoyahav, line 333: Should this be using GetValues? (similarly below)
Also, accessing the first element here seems a bit arbitrary, should we assert that it's non-empty and add a comment about the splits being the same for all? (move the comment from below up)

tfx/components/transform/executor.py
```diff
@@ -329,10 +329,20 @@ def Do(self, input_dict: Dict[Text, List[types.Artifact]],
...
+
+      # Verify that all input artifacts have the same set of split names.
+      split_names_set = set(split_names)
+      for artifact in input_dict[EXAMPLES_KEY]:
```
zoyahav, line 337: Could you make this a list comprehension so that we're iterating the artifact_split_names?

tfx/components/transform/executor_test.py
```diff
@@ -58,9 +60,25 @@ class _TempPath(types.Artifact):
 # TODO(b/122478841): Add more detailed tests.
 class ExecutorTest(tft_unit.TransformTestCase):
 
-  def _get_source_data_dir(self):
-    return os.path.join(
+  @classmethod
+  def setUpClass(cls):
+    cls._temp_example_dir = tempfile.mkdtemp()
```
zoyahav, line 65: Could you move all of these cls._... = ... to just be top level in the class?
So only the things that can't be static such as copy_dir, glob, etc. are done here.
Also, note that class level constants should be named with all caps.

tfx/components/transform/executor_test.py
```diff
@@ -58,9 +60,25 @@ class _TempPath(types.Artifact):
...
+    source_example_dir = os.path.join(cls._source_data_dir, 'csv_example_gen')
+
+    cls._artifact1_uri = os.path.join(cls._temp_example_dir, 'csv_example_gen1')
+    io_utils.copy_dir(source_example_dir, cls._artifact1_uri)
+    cls._artifact2_uri = os.path.join(cls._temp_example_dir, 'csv_example_gen2')
+    io_utils.copy_dir(source_example_dir, cls._artifact2_uri)
+
+    # Duplicate the number of train and eval records such that
+    # second artifact has twice as many as first.
+    artifact2_pattern = os.path.join(cls._artifact2_uri, '*', '*')
```
zoyahav, line 77: Can this just be replaced with join(..., '**')?
(depends on the file structure, so I'm not sure)

tfx/components/transform/executor_test.py
```diff
@@ -72,24 +90,35 @@ def _get_output_data_dir(self, sub_dir=None):
...
-    examples = standard_artifacts.Examples()
-    examples.uri = os.path.join(source_data_dir, 'csv_example_gen')
-    examples.split_names = artifact_utils.encode_split_names(['train', 'eval'])
+    e1 = standard_artifacts.Examples()
+    e1.uri = self._artifact1_uri
+    e1.split_names = artifact_utils.encode_split_names(['train', 'eval'])
+    e2 = copy.deepcopy(e1)
+    e2.uri = self._artifact2_uri
+
+    self._single_artifact = [e1]
```
zoyahav, line 99: Instead of having single_artifact vs. multiple_artifacts, can we just have artifacts that the tests can choose to just use the first for example?

tfx/components/transform/executor_test.py
```diff
@@ -72,24 +90,35 @@ def _get_output_data_dir(self, sub_dir=None):
...
     }
 
     # Create output dict.
     self._transformed_output = standard_artifacts.TransformGraph()
     self._transformed_output.uri = os.path.join(output_data_dir,
                                                 'transformed_graph')
-    self._transformed_examples = standard_artifacts.Examples()
-    self._transformed_examples.uri = os.path.join(output_data_dir,
-                                                  'transformed_examples')
+    te1 = standard_artifacts.Examples()
```
zoyahav, line 114: Let's avoid these types of abbreviations for variables (transformed_1 for example is fine)

tfx/components/transform/executor_test.py
```diff
@@ -72,24 +90,35 @@ def _get_output_data_dir(self, sub_dir=None):
...
+    te1.uri = os.path.join(output_data_dir, 'transformed_examples', '1')
+    te2 = standard_artifacts.Examples()
+    te2.uri = os.path.join(output_data_dir, 'transformed_examples', '2')
+
+    self._single_transform = [te1]
+    self._multiple_transform = [te1, te2]
```
zoyahav, line 120: Add _examples suffix for a class attribute

tfx/components/transform/executor_test.py
```diff
@@ -127,7 +154,8 @@ def setUp(self):
     # Executor for test.
     self._transform_executor = executor.Executor()
 
-  def _verify_transform_outputs(self, materialize=True, store_cache=True):
+  def _verify_transform_outputs(self, materialize=True, store_cache=True,
+                                multiple_transform=False):
```
zoyahav, line 158: multiple_example_inputs?

tfx/components/transform/executor_test.py
```diff
@@ -136,23 +164,51 @@ def _verify_transform_outputs(self, materialize=True, store_cache=True):
...
-      eval_pattern = os.path.join(self._transformed_examples.uri, 'eval', '*')
-      eval_files = tf.io.gfile.glob(eval_pattern)
-      self.assertNotEqual(0, len(eval_files))
-      eval_dataset = tf.data.TFRecordDataset(
-          eval_files, compression_type='GZIP')
-
-      self.assertGreater(
-          _get_dataset_size(train_dataset), _get_dataset_size(eval_dataset))
+      for example, transformed_example in zip(
+          example_artifacts, transformed_example_artifacts):
```
zoyahav, line 176: Assert the lengths before zipping

tfx/components/transform/executor_test.py
```diff
@@ -136,23 +164,51 @@ def _verify_transform_outputs(self, materialize=True, store_cache=True):
...
+        examples_train_pattern = os.path.join(example.uri, 'train', '*')
```
zoyahav, line 177: No need for this temp variable, similarly below

tfx/components/transform/executor_test.py
```diff
@@ -136,23 +164,51 @@ def _verify_transform_outputs(self, materialize=True, store_cache=True):
...
+        examples_train_files = tf.io.gfile.glob(examples_train_pattern)
+        transformed_train_pattern = os.path.join(transformed_example.uri,
+                                                 'train', '*')
+        transformed_train_files = tf.io.gfile.glob(transformed_train_pattern)
+        self.assertNotEqual(0, len(transformed_train_files))
```
zoyahav, line 182: This can just be assertTrue, similarly below

tfx/types/artifact_utils.py
```diff
@@ -98,7 +98,12 @@ def get_split_uris(artifact_list: List[Artifact], split: Text) -> List[Text]:
     split_names = decode_split_names(artifact.split_names)
     if split in split_names:
       matching_artifacts.append(artifact)
-  return [os.path.join(artifact.uri, split) for artifact in matching_artifacts]
+  split_uris = [os.path.join(artifact.uri, split)
+                for artifact in matching_artifacts]
+  if len(split_uris) != len(artifact_list):
+    raise ValueError(
```
zoyahav, line 104: Add raises to the docstring?

tfx/types/artifact_utils.py
```diff
@@ -98,7 +98,12 @@ def get_split_uris(artifact_list: List[Artifact], split: Text) -> List[Text]:
     split_names = decode_split_names(artifact.split_names)
     if split in split_names:
       matching_artifacts.append(artifact)
```
zoyahav, line 100: We could put the os.path.join() call here and rename matching_artifacts to result (return it in the end)

zoyahav (approved):
Thanks!

tfx/components/transform/executor_test.py
18jeffreyma: i think double-asterisk matches recursively (i.e. matches files at both 1 level depth and 2-level depth) when we only want to match files at the 2nd depth

18jeffreyma:
addressed latest round of feedback, thanks!

tfx/types/artifact_utils.py
```diff
@@ -92,13 +92,20 @@ def get_split_uris(artifact_list: List[Artifact], split: Text) -> List[Text]:
...
+      number of input artifacts.
   """
-  matching_artifacts = []
+  result = []
   for artifact in artifact_list:
     split_names = decode_split_names(artifact.split_names)
     if split in split_names:
-      matching_artifacts.append(artifact)
-  return [os.path.join(artifact.uri, split) for artifact in matching_artifacts]
+      result.append(os.path.join(artifact.uri, split))
```
ruoyu90, line 104: Seems we're leaking the details of split uri assembling. Shall we use / enhance existing utility function to handle this? e.g., `artifact_utils.get_split_uris()`.

ruoyu90 (commented):
Thanks Jeffrey! Overall LGTM modulo one small comment.

tfx/types/artifact_utils.py
18jeffreyma: spoke about this at Wednesday sync, marked as resolved.

ruoyu90 (approved):

1025KB (approved):

ConverJens:
@18jeffreyma @1025KB @ruoyu90  Is this confirmed to work? I'm using a resolver to feed transform (with N latest example artifacts) and for me, this fails with an assertion error:

`Traceback (most recent call last):
  File "/tfx-src/tfx/orchestration/kubeflow/container_entrypoint.py", line 360, in <module>
    main()
  File "/tfx-src/tfx/orchestration/kubeflow/container_entrypoint.py", line 353, in main
    execution_info = launcher.launch()
  File "/tfx-src/tfx/orchestration/launcher/base_component_launcher.py", line 205, in launch
    execution_decision.exec_properties)
  File "/tfx-src/tfx/orchestration/launcher/in_process_component_launcher.py", line 67, in _run_executor
    executor.Do(input_dict, output_dict, exec_properties)
  File "/pipeline/kubeflow/custom_components/transform_master/executor.py", line 425, in Do
    self.Transform(label_inputs, label_outputs, status_file)
  File "/pipeline/kubeflow/custom_components/transform_master/executor.py", line 911, in Transform
    materialize_output_paths)
  File "/pipeline/kubeflow/custom_components/transform_master/executor.py", line 1309, in _MakeDatasetList
    assert len(file_patterns) == len(materialize_output_paths)
AssertionError
`

ConverJens:
@18jeffreyma @1025KB @ruoyu90  I might be mistaken here, but depending on how the multiple inputs/outputs is supposed to work  I think the test for multiple artifacts might be flawed.

The test case defines the input examples AND the output examples explicitly, i.e. two examples in and two out. However, in a real-world scenario, it is more likely that the number of outputs needs to be dynamically determined from the number of inputs.

From the test cases and your discussions, it seems as Transform is supposed to register multiple artifacts in MLMD and if that  is the case, then the outputs directories needs to determined dynamically from the URIs provided by MLMD.

Please, let me know if I have misunderstood the functionality or if this logic is indeed missing.

Issue #2380, opened 2020-08-25
Add MLMD read client for exploration from interactive environments


casassg:
As a user I want to load pipeline status (and see what components are being executed) from an interactive context.

Even though this is possible at the moment, it requires user to have a deep understanding of MLMD in order to get status of pipeline runs. It would be great to have something like this:

```py
mlmd = MetadataClient(connection_config)
pipeline = mlmd.get_pipeline('pipeline_name')
>>> pipeline.runs = [PipelineRun('pipeline_name', '2020-08-24T23:31:03.286137'), PipelineRun('pipeline_name', '2020-08-21T22:42:25.102927')]

components = pipeline.runs[-1].components
>>> components = [Component('BigQueryExampleGen', 'complete'), Component('StatsGen', 'new')]# we can probs make this a more verbose name or keep as is.
```

We could even add something like InteractiveContext to explore results. 
```py
components[1].show_output('stats') # Loads viz for TFDV

# or

components[0].load_dataset() # Returns a tf.data.Dataset instance for interactive exploration 
```

This would make the task of interactively exploring results from pipeline runs easier for the user. Currently working on something similar to this as part of a remote job scheduler from notebook environments (aka schedule a pipeline in an adhoc job from a notebook). 

Code snippets are just suggestions on API, would love to discuss those as well.

Existing alternatives and why they are not fully great:
- Airflow workshop TFX utils: https://github.com/tensorflow/tfx/tree/master/tfx/examples/airflow_workshop/notebooks - Assumes pipeline has its own MLMD instance separately. Doesn't work if you start using a shared MLMD instance.
- Orchestration Metadata: https://github.com/tensorflow/tfx/blob/master/tfx/orchestration/metadata.py - I started developing something based on this. Could PR new functions to this for usage. But doesn't have easy to use functions to query status as I show above.

zhitaoli:
HI @casassg,

I like the idea you proposed. Something similarly was also mentioned by @hongye-sun once. 

@ruoyu90 Can you help @casassg on this?

Some replies inlined:

> It would be great to have something like this:

WDYT on building this directly in TFX side, possibly in the [Pipeline](https://github.com/tensorflow/tfx/blob/5a4229639343a373fe4da9295b1ee5113912a902/tfx/orchestration/pipeline.py#L57) object? `Run` is a concept we modeled on top of MLMD, instead of provided by latter.

> We could even add something like InteractiveContext to explore results.

[...]

PR #2415, opened 2020-08-29
[Proposal] MLMD Client for TFX Pipelines

casassg:
Related to #2380 

Proposing to add a new MLMD read client that allows us to read status from Pipeline components on each run. This is a small proposal to start adding new functionalities in the future. 

--- 
Unfortunately I have not been able to setup bazel locally for this project which would allow me to run tests before hand.


casassg:
@zhitaoli should we reopen this?

PR #2428, opened 2020-09-02
Testing Trainer and Transform support for multiple artifact input and output channels with LatestArtifactResolver

18jeffreyma:
Testing Trainer and Transform component support for multiple artifact input and output using latest artifact resolver. To overview, this e2e example runs a simple ExampleGen pipeline to ingest several spans of data, then uses ResolverNode to retrieve the latest   few examples in a single output channel, pass that output channel to the Transform component, then pass that multi-artifact output channel to Trainer for training.

TODO(jjma): Add TransformCache testing to this E2E PR once we've finalized discussion on Resolver behavior. Current concerns are that Transform output examples could be retrieved by ResolverNode (since it retrieves based on execution id and not span specifically yet).


18jeffreyma:
@1025KB currently in draft form, but looking at ways to add some testing for TransformCache as well. Currently still concerned about Transform output also being resolved by ResolverNode, so will wait until our sync to add those in.

18jeffreyma:
Verified this behavior, marking as closed.

PR #2438, opened 2020-09-03
Correct Trainer guide doc to use base_model

rclough:
Since 0.2X, the Trainer Guide documentation has referenced `base_models` (plural), but this does not match the component spec/executor/what exists in `trainer_fn_args` (`base_model` (singular)); in effect, the example code does not work. 

This PR address it.


rclough:
@googlebot  I signed it!

rclough:
@googlebot I signed it!

rclough:
This should be reopened, the docs are still wrong...

PR #2439, opened 2020-09-03
Adding Span-based resolver with RangeConfig support to retrieve windows of span

18jeffreyma:
NOTE: This PR is dependent on [PR#2339](https://github.com/tensorflow/tfx/pull/2339), which introduces the RangeConfig proto. That PR must be merged before this one can be merged.

Introduced a new Resolver, SpansResolver, which uses a provided RangeConfig (using either static range or rolling range), and returns an input Channel containing the resolved artifacts. Previously, LatestArtifactResolver resolved artifacts based on execution id, which does not map to span properly (i.e. a earlier Span can have a newer execution id, if we decide to process a new version of a Span). Thus, adding a Span based resolver is necessary to support correct retrieval of moving windows of Span (as proposed in the continuous pipeline architecture).


18jeffreyma:
@1025KB

tfx/dsl/experimental/spans_resolver.py
```diff
@@ -0,0 +1,117 @@
...
+        latest_artifact = previous_artifacts[0]
+        artifacts_dict[k] = [
+            artifact_utils.deserialize_artifact(
+                latest_artifact.type, latest_artifact.artifact)
+        ]
+        resolve_state_dict[k] = True
+
+    return base_resolver.ResolveResult(
+        per_key_resolve_result=artifacts_dict,
+        per_key_resolve_state=resolve_state_dict)
```
1025KB, line 117: I'm thinking even we have multiple examplegen as input source, but we only want to output a single example artifact channel for trainer

e.g., a for loop of examplegens that process span 3,4,5 and a resolver that resolves a channel of example artifact (3,4,5).

the artifacts_dict[k], can k be the type instead of name?

_another option (need to verify if it works or not) is using [this](https://github.com/tensorflow/tfx/blob/master/tfx/components/base/base_node.py#L153) for resolver to add examplegen dependency, and resolver just resolve based on component type (not id) and artifact type_

tfx/proto/range_config.proto
```diff
@@ -0,0 +1,59 @@
+// Copyright 2020 Google LLC. All Rights Reserved.
```
1025KB, line 1: just confirm, this pr doesn't depends on rangeconfig pr a lot other than this proto config, right?

18jeffreyma: yea it only depends on range_config PR for the proto.

tfx/dsl/experimental/spans_resolver.py
18jeffreyma: i see what you mean (same as the paste in the doc right?). the design for this resolver is more following the logic for the LatestArtifactResolver, where its possible for a resolver to resolve multiple channels. are we fixing this resolver to only resolver one channel always? if so, we can definitely make k be the type name and instead of creating an output for each source channel, we can just aggregate them all into one output.

1025KB: for our use case, we still want the artifact comes from certain component (e.g., example artifact comes from examplegens, not transforms), 

if we just have one examplegen in the pipeline, current code would be fine.

but if we config multiple examplegen in the pipeline, we need a way to get a channel for all examplegen's output artifacts

maybe add a boolean merge_same_artifact_type or sth, if set to true, same type will be treated as a single (the 'a' below, not 'b')

a.multiple channel with same type -> merge -> process -> output single channel
b.multiple channel with same type -> process -> merge -> output single channel


_for the other option I mentioned above, channel need to contains the component type info besides component id info_

18jeffreyma: updated PR to merge multiple channels if flag is set. it merges these into a resolved output with key thats the same as the type name `Examples`.

18jeffreyma:
/cc @hughmiao 
/cc @ruoyu90 
for reference, updated PR to resolve latest span, but also include functionality to merge channels into one output Example artifact channel.

tfx/dsl/experimental/spans_resolver.py
```diff
@@ -0,0 +1,158 @@
...
+    resolve_state_dict = {}
+
+    # Verifies that duplicate spans are not added to output over multiple
+    # channels; specifically for when `merge_same_artifact_type` is set to True.
+    total_processed_spans = set()
+
+    pipeline_context = metadata_handler.get_pipeline_context(pipeline_info)
+    if pipeline_context is None:
+      raise RuntimeError('Pipeline context absent for %s' % pipeline_context)
+    for k, c in source_channels.items():
```
ruoyu90, line 62: Do we intend to allow multiple channels for this specific resolver? Or we only expect one?

tfx/dsl/experimental/spans_resolver.py
ruoyu90: Span is a special concept that only applies for `Examples`. Combining multiple `Channel`s into one can be handled separately. Shall we just assert that this resolver takes exactly one source which is of type `Examples`? This will also make migration to the soon-to-be-there new stack easier. This will also simplify the logic a bit.

18jeffreyma: i think from a design standpoint it might make more sense. are you proposing having a 2nd resolver that can resolve multiple channels of input into a single channel of output?

tfx/dsl/experimental/spans_resolver.py
18jeffreyma: currently, we're allowing multiple channels since current plan is to also add ability to resolve multiple channels into one. in the event, we decide to move resolver in different direction, we can add the one channel assertion.

tfx/dsl/experimental/spans_resolver.py
18jeffreyma: from discussion with hui and ruoyu today, this resolver should mainly be for taking one input channel (of Examples) and retrieving the range of spans. likely a different primitive will handle the case where we want to resolve multiple example_gens into a single channel, so to make the logic cleaner, we will simply force this resolver to only conssider one channel.

tfx/dsl/experimental/spans_resolver.py
18jeffreyma: re todays  discussion, we will assert only one channel is allowed and resolve that single Examples channel.

18jeffreyma:
updated resolver behavior. new span resolver assumes only one source channel as input, then resolves over that channel. from discusion from ruoyu and hui, the behavior to merge output from multiple components will be handled by a different primitive (rather than being in resolver).

ConverJens:
This seems like great and needed functionality! What's the status on this?

1025KB:
> This seems like great and needed functionality! What's the status on this?

Current Resolver can resolve latest n spans, and this is focus on resolving a set of user specified spans, it's current WIP

May I know your use cases to help us make decision related to what interface we provide to users?

ConverJens:
@1025KB Right, I might have misunderstood the functionality then.
The use case would be to retrieve a rolling window of spans from date directories. Given data structured in directories on the form:
20200928
20200929
20200930
20201001
20201002

run a pipeline on the current date and X days back. 

Is this already supported?

ConverJens:
@1025KB Will the current Resolver fetch the latest spans if one of the days has been reprocessed? E.g. in case of bad data from upstream, and perhaps a new version of data has been added.

1025KB:
note that current resolver will get the latest based on processed time, it's not span based (which should be fixed with the recent span and version spec support in ExampleGen)

Note that if you just need one span for training, you don't need resolver. [ExampleGen](https://github.com/tensorflow/tfx/blob/master/docs/guide/examplegen.md#version) will pickup the latest version of a specified span, and trainer can just use ExampleGen's output

1025KB:
> @1025KB Right, I might have misunderstood the functionality then.
> The use case would be to retrieve a rolling window of spans from date directories. Given data structured in directories on the form:
> 20200928
> 20200929
> 20200930
> 20201001
> 20201002
> 
> run a pipeline on the current date and X days back.
> 
> Is this already supported?

we plan to provide a RFC/story about periodical training, but it's still WIP.

the idea is like examplegen process the new data every run, and resolver get the window of span for training, e.g.,
 
pipeline run 5 generates span 5 in ExampleGen, Trainer gets span 3 4 5 from resolver
pipeline run 6 generates span 6 in ExampleGen, Trainer gets span 4 5 6 from resolver

ConverJens:
> > @1025KB Right, I might have misunderstood the functionality then.
> > The use case would be to retrieve a rolling window of spans from date directories. Given data structured in directories on the form:
> > 20200928
> > 20200929
> > 20200930
> > 20201001
> > 20201002
> > run a pipeline on the current date and X days back.
> > Is this already supported?
> 
> we plan to provide a RFC/story about periodical training, but it's still WIP.
> 
> the idea is like examplegen process the new data every run, and resolver get the window of span for training, e.g.,
> 
> pipeline run 5 generates span 5 in ExampleGen, Trainer gets span 3 4 5 from resolver
> pipeline run 6 generates span 6 in ExampleGen, Trainer gets span 4 5 6 from resolver

This is exactly what I'm looking for.

ConverJens:
> note that current resolver will get the latest based on processed time, it's not span based 

And I guess that this is the issue with the current approach, if any span is reprocesses it might provide me with data that I didn't intend. 

> (which should be fixed with the recent span and version spec support in ExampleGen)

What do you mean by this? I've seen the latest support for dates as spans but the range_config proto seems to only be present in master still, not 0.24.0, and even then it's only a static config which only supports choosing a specific span, not a range as far as I could tell.

> Note that if you just need one span for training, you don't need resolver. [ExampleGen](https://github.com/tensorflow/tfx/blob/master/docs/guide/examplegen.md#version) will pickup the latest version of a specified span, and trainer can just use ExampleGen's output

I'm currently running the pipeline like this but only for testing purposes. Otherwise my only option is to duplicate data everytime I want to include a new data which is not feasible for production.

ConverJens:
@1025KB @18jeffreyma Any update on this?

1025KB:
> @1025KB
> 
> > note that current resolver will get the latest based on processed time, it's not span based
> 
> And I guess that this is the issue with the current approach, if any span is reprocesses it might provide me with data that I didn't intend.
> 
> > (which should be fixed with the recent span and version spec support in ExampleGen)
> 
> What do you mean by this? I've seen the latest support for dates as spans but the range_config proto seems to only be present in master still, not 0.24.0, and even then it's only a static config which only supports choosing a specific span, not a range as far as I could tell.
> 

ExampleGen process one span at a time, but we do plan to support range of span for resolver (as this PR)

> > Note that if you just need one span for training, you don't need resolver. [ExampleGen](https://github.com/tensorflow/tfx/blob/master/docs/guide/examplegen.md#version) will pickup the latest version of a specified span, and trainer can just use ExampleGen's output
> 
> I'm currently running the pipeline like this but only for testing purposes. Otherwise my only option is to duplicate data everytime I want to include a new data which is not feasible for production.

1025KB:
> @1025KB @18jeffreyma Any update on this?

Sorry for the late updates, we are working on this Q4

ConverJens:
> > @1025KB @18jeffreyma Any update on this?
> 
> Sorry for the late updates, we are working on this Q4

Thanks for the information!

ConverJens:
@1025KB Is there also any logic for allowing components (StatsGen, Transform, Trainer etc.) to consume a channel with multiple example artifacts, like the output of a resolver? From my attempts, they always assume that only one example artifact is present in the channel.

ConverJens:
@1025KB I have made a slightly generalized StatisticsGen which can consume multiple artifacts from a single channel, would you like me to contribute it?

PR #2606, opened 2020-10-06
Add test to remove fields from artifact

casassg:
In https://github.com/tensorflow/tfx/issues/2506 we have detected an issue when running BigQueryExampleGen in different versions of TFX. After reviewing some code I wanted to check if the issue was in registering Artifacts where fields have been removed. It seems so. 

This PR will be updated with a fix. Current added test fails to run which allows us to figure out a solution.

### DIFF
--- a/tfx/orchestration/metadata.py
+++ b/tfx/orchestration/metadata.py
@@ -182,7 +182,7 @@ class Metadata(object):
     if artifact_type.id:
       return artifact_type
     type_id = self.store.put_artifact_type(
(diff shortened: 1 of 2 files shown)


casassg:
Seems fix is not possible due to this not being implemented in MLMD side: https://github.com/google/ml-metadata/blob/1fceefd596e2ed7e98e629e7186e0f0b1fbace28/ml_metadata/metadata_store/metadata_store.cc#L248

PR #2636, opened 2020-10-11
Bumping Beam version to use 2.23.0 for TFX 0.21.5 Patch Release

dhruvesh09:


zhitaoli (changes requested):
I did not realize that for r0.21 it's already upper bounded. I don't think we need this change.

PR #3300, opened 2021-02-26
Add configuration option for project used in BQ pusher

casassg:
Add a way to configure what project the query is run for BQ Pusher. This allows for more flexible configuration without requiring to set an environment variable.


casassg:
Closed by dbd4d1e

PR #3454, opened 2021-03-25
Fixes typos in taxi_utils.py

numerology:
nit fixes.


numerology:
@googlebot I signed it!

PR #3544, opened 2021-04-14
Make ExampleGen driver fingerprint execution property optional

aoen:
Make ExampleGen driver fingerprint execution property optional. This is necessary since otherwise the QueryBased Driver created by @1025KB in
https://github.com/tensorflow/tfx/commit/8b6e3f12a0b3015ea3a02115abf6a18ce92400b8 does not work and raises an exception (since it does not generate a fingerprint).

### DIFF
--- a/tfx/components/example_gen/driver.py
+++ b/tfx/components/example_gen/driver.py
@@ -53,9 +53,10 @@ def update_output_artifact(
     exec_properties: execution properties passed to the example gen.
     output_artifact: the example artifact to be output.
   """
-  output_artifact.custom_properties[
-      utils.FINGERPRINT_PROPERTY_NAME].string_value = str(
-          exec_properties[utils.FINGERPRINT_PROPERTY_NAME])
+  if exec_properties[utils.FINGERPRINT_PROPERTY_NAME] is not None:
+    output_artifact.custom_properties[
+        utils.FINGERPRINT_PROPERTY_NAME].string_value = (
+            exec_properties[utils.FINGERPRINT_PROPERTY_NAME])
   output_artifact.custom_properties[
       utils.SPAN_PROPERTY_NAME].string_value = str(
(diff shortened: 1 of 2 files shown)


aoen:
@googlebot I signed it!
Added myself to my company's (Twitter) Google Group which I believe should pass the check now, let me know if it does not.

tfx/components/example_gen/driver.py
```diff
@@ -53,9 +53,10 @@ def update_output_artifact(
     exec_properties: execution properties passed to the example gen.
     output_artifact: the example artifact to be output.
   """
-  output_artifact.custom_properties[
-      utils.FINGERPRINT_PROPERTY_NAME].string_value = str(
-          exec_properties[utils.FINGERPRINT_PROPERTY_NAME])
+  if exec_properties[utils.FINGERPRINT_PROPERTY_NAME] is not None:
```
zhitaoli, line 56: I think we should have a unit test in https://github.com/tensorflow/tfx/blob/master/tfx/components/example_gen/driver_test.py when this is None (or unset)

Do you think you can add that?

zhitaoli, line 56: Should we exec_properties.get(utils.FINGERPRINT_PROPERTY_NAME) to guard against the key not in this exec_properties? Or that guaranteed by an API contract?

@1025KB

zhitaoli (changes requested):
Thanks! I think we need a unit test for this one.

tfx/components/example_gen/driver.py
1025KB: For official drivers, fingerprint will also be set (but we might set it to None when fingerprint is not available)

aoen: Done, added tests for the other properties while I was at it.

zhitaoli (approved):

Issue #3573, opened 2021-04-20
Penguin Example -> AttributeError: module 'user_module_0' has no attribute 'trainer_fn'


RossKohler:
**System information**

- Have I specified the code to reproduce the issue(Yes/No): Yes
- Environment in which the code is executed (e.g., Local(Linux/MacOS/Windows),
  Interactive Notebook, Google Cloud, etc): MacOS
- TensorFlow version (you are using): ^2.4.1
- TFX Version: 0.28.0
- Python version: 3.7.0

**Describe the current behavior**

Cloning the TFX repo (latest -> ce2fd6fbde9845cd837c47089c3d6db2f87007b9) and attempting to run the penguin example using the LocalDAGRunner at `penguin_pipeline_local.py` yields the following error:

```
AttributeError: module 'user_module_0' has no attribute 'trainer_fn'
```

**Describe the expected behavior**

Running the penguin example using the localDAGRunner completes successfully.

**Steps to resolve issue**

I believe there is a line missing in the instantiation of the `Trainer` component in the pipeline. An additional parameter of `custom_executor_spec=executor_spec.ExecutorClassSpec(GenericExecutor)` should be added. This fixes the issue.

RossKohler:
I've created a PR to address this issue [here](https://github.com/tensorflow/tfx/pull/3574)

jiyongjung0:
Thank you for reporting.

As I responded in the PR, we are currently preparing our next release 0.30.0, and the examples were changed a bit earlier. You can add cusotm_executor_spec by yourself, or use the nightly version until 0.30.0 release finalized.

Thanks!

PR #3574, opened 2021-04-20
Fixes issue https://github.com/tensorflow/tfx/issues/3573

RossKohler:
See issue [here](https://github.com/tensorflow/tfx/issues/3573)


RossKohler:
@googlebot I signed it!

jiyongjung0:
Thank you so much for your contribution, and your change is right. We should have custom_executor configs.

We are currently preparing our next release 0.30.0, and it seems like the example changes went ahead of the final release.
The parameter will not be needed any more after 0.30.0, so the config was deleted in 
https://github.com/tensorflow/tfx/commit/b600bc14739456bc1678a0d1f0adab7a3cfac931 .
So I'm a bit hesitant to merge (which is basically a rollback) at this moment.

RossKohler:
No worries :) I'll close the PR

PR #3705, opened 2021-05-10
TFX 0.26.4 Patch Release

dhruvesh09:


jiyongjung0 (commented):
Thank you for preparing this patch release!

Could you update tfdv dependency in dependencies.py as well? As we described in the release note.

zhitaoli (approved):

PR #4018, opened 2021-07-12
Include uuid4 string in default generated ai platform job name.

johnPertoft:
In a TFX pipeline with multiple parallel ai platform jobs usually all but one will fail to start since the default job id will be the same for all jobs. This PR adds a substring of a uuid4 to avoid the naming collision.

### DIFF
--- a/tfx/extensions/google_cloud_ai_platform/training_clients.py
+++ b/tfx/extensions/google_cloud_ai_platform/training_clients.py
@@ -17,6 +17,7 @@ import abc
 import datetime
 import json
 from typing import Any, Dict, List, Optional, Text, Union
+import uuid
 
 from absl import logging
 from google.cloud.aiplatform import gapic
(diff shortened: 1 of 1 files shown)


johnPertoft:
@googlebot I signed it!

charlesccychen:
@1025KB @neuromage

Do we see any issues with adding this additional 8 character suffix?

neuromage:
Should be fine I think, LGTM.

charlesccychen (approved):

charlesccychen:
We will use `'%08x' % random.getrandbits(32)` instead of uuid because several bits in the first 8 digits of the UUID format are predetermined (e.g. the versioning and variant fields; https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)).

charlesccychen:
We will use `'%08x' % random.getrandbits(32)` instead of uuid because several bits in the first 8 digits of the UUID format are predetermined (e.g. the versioning and variant fields; https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)).

tfx/extensions/google_cloud_ai_platform/training_clients.py
```diff
@@ -423,10 +424,11 @@ def create_training_args(self, input_dict: Dict[Text, List[types.Artifact]],
         {telemetry_utils.LABEL_TFX_EXECUTOR: executor_class_path}):
       job_labels = telemetry_utils.make_labels_dict()
 
-    # 'tfx_YYYYmmddHHMMSS' is the default job display name if not explicitly
-    # specified.
-    job_id = job_id or 'tfx_{}'.format(
-        datetime.datetime.now().strftime('%Y%m%d%H%M%S'))
+    # 'tfx_YYYYmmddHHMMSS_xxxxxxxx' is the default job display name if not
+    # explicitly specified.
+    job_id = job_id or 'tfx_{}_{}'.format(
```
zhitaoli, line 429: Is newly generated job_id logged anywhere later? If not we should use absl.logging.info since the random number is not inferrable elsewhere.

johnPertoft:
Is there still an interest in this PR? I saw that recent versions included support for the vertex ai jobs as well and I haven't verified if they still suffer from the same problem. I'm also not using tfx at the moment so it's not a pressing issue for me.

charlesccychen:
Yes, thank you @johnPertoft -- I approved it in an internal system that should have pushed this through, but it looks like it got stuck on an internal test. I will try to get this through later today.

PR #4082, opened 2021-07-23
Bump Kubernetes dep up

casassg:
We need to use kubernetes~=12 to use `refresh_api_key_hook`  for kubernetes authentication.  Due to TFX we cant get a full closure with the new k8s version.

If we can cut a release for this under 1.0.1 that would be great!


casassg:
CC @zhitaoli

casassg:
Also CC @rcrowe-google

jiyongjung0 (approved):
Thanks!

PR #4127, opened 2021-08-06
Propose using separate function to generate container command

casassg:
Container command can actually be used outside of AIPT. We would find this useful to write our own abstractions on top of the existing training client, and avoid having to generate this container command again.


casassg:
CC @zhitaoli as the latest code reviewer of the file

.github/workflows/ci-test.yml
```diff
@@ -42,10 +42,22 @@ jobs:
...
+    - name: Cache pip
+      uses: actions/cache@v2
+      with:
+        # This path is specific to Ubuntu
+        path: ~/.cache/pip
+        # Look to see if there is a cache hit for the corresponding setup.py + TFX version 
+        key: ${{ runner.os }}-pip-${{ hashFiles('tfx/dependencies.py') }}-
+        restore-keys: |
+          ${{ runner.os }}-pip-${{ hashFiles('tfx/dependencies.py') }}-
+          ${{ runner.os }}-pip-
```
jiyongjung0, line 54: Out of curiosity. Why do we need this second restore-key?

And it seems like we don't need to repeat the one defined in "key" in restore-key. (https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#using-the-cache-action)

tfx/extensions/google_cloud_ai_platform/training_clients.py
```diff
@@ -112,14 +112,14 @@ def _create_job_spec(
   @abc.abstractmethod
   def launch_job(self,
                  job_id: Text,
-                 parent: Text,
+                 project: Text,
```
jiyongjung0, line 115: Thank you for the fix!

tfx/extensions/google_cloud_ai_platform/training_clients.py
```diff
@@ -149,6 +149,35 @@ def get_job_name(self) -> Text:
     """Gets the job name."""
     return self._job_name
 
+  def generate_container_command(self,
```
jiyongjung0, line 152: How about adding an underscore to mark this method  as a protected method like '_generate_container_command()' if this method will be called only inside the class (or subclasses)?

https://google.github.io/styleguide/pyguide.html#3162-naming-conventions

jiyongjung0 (commented):
Thank you so much for the cl and I left a few minor comments.

I didn't aware of the caching in actions and it looks great!!

.github/workflows/ci-test.yml
casassg: Second key is used to restore from cache a partial hit (for example if changes in deps files, it will retrieve the previous cache). This allows to reuse at least the shared deps wheels even if dependencies have changed.

tfx/extensions/google_cloud_ai_platform/training_clients.py
casassg: pylint was complaining to me :D

tfx/extensions/google_cloud_ai_platform/training_clients.py
casassg: So, the reason I made it public was so that inheriting classes outside of TFX can reuse it. The idea is that at Twitter we are building a similar Trainer using KF TFJob instead. With this method public, we can just inherit the client and generate the container command without having to vendor the code.

Given our class will be in an inherited class, then theoretically, you should not use protected or private methods. 

Also, making it public ensures a bit more of stability for us to derive from and the cost is low due to low complexity of underlying code.

Let me know if that's okay, otherwise I can make it private.

.github/workflows/ci-test.yml
casassg: (also removed the first key)

jiyongjung0: Thank you for the explanation! I should have pay more attention on the documentation.... It seems great!

jiyongjung0 (approved):
Thanks!

casassg:
Any chance we can get this included in the upcoming release?

casassg:
@jiyongjung0 bringing this back up!

jiyongjung0:
I'm sorry but I merged your PR with the previous state and this PR was not marked as merged as usual due to a difference in code style.

I think that all the required content were added. Please reopen or let me know if you find any issues. Thank you for contribution!

casassg:
looks good, ty!

PR #4144, opened 2021-08-13
Prototype remote @component executor

casassg:
Still under work, but the idea here is to have a way to define @component components without having to send the modules to KFP/Vertex and instead serializing from the client on pipeline compilation time.

Need to:
- [x] Serialize executor
- [x] Serialize component
- [ ] Make it optional to use this version

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -139,6 +139,7 @@ dmypy.json
 # Bazel generated files
 bazel-*
 **/*_pb2.py
+**/*_pb2_grpc.py
 # LINT.ThenChange(.gitignore)
 
 # Github integration settings
--- a/.gitignore
+++ b/.gitignore
@@ -139,4 +139,5 @@ dmypy.json
 # Bazel generated files
 bazel-*
 **/*_pb2.py
+**/*_pb2_grpc.py
 # LINT.ThenChange(.dockerignore)
--- a/tfx/dependencies.py
+++ b/tfx/dependencies.py
@@ -60,6 +60,7 @@ def make_pipeline_sdk_required_install_packages():
           git_master='@git+https://github.com/google/ml-metadata@master'),
       'packaging>=20,<21',
       'portpicker>=1.3.1,<2',
+      'strip-hints>=0.1.0<1'
       'protobuf>=3.13,<4',
       'docker>=4.1,<5',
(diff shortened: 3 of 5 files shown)


casassg:
@charlesccychen @zhitaoli wondering if we could apply something like this (I believe this may still not work as it still requires the component to be accessible on the python path, but need to validate it)

Also this is a backwards incompatible change as imports need to be specified inside the function or else it wont really work due to the isolated nature of exec

charlesccychen:
Hey @casassg, is the intention to create something similar to the "KFP light-weight Python component" functionality, where the component is run inside a specified container directly from the entrypoint? Is this to avoid container building, and will the container contain the TFX SDK?

casassg:
Yes, the idea is to replicate KFP lightweight component. The idea is also to make it generic enough so it doesnt require container to run and can be used from LocalDagRunner, KubeflowDagRunner (and v2) and more 

In the case of KFP indeed it would be run as a container since all components are run individually.

Main issue I see w custom components currently is that they require a custom docker image to be built in order to send the component module to the KFP runtime. Instead by serializing most of it and just executing, it allows us to define custom components and send to KFP without docker image (similar to the lightweight component indeed).

Main difference w KFP lightweight is that this can be used in the current python context (while at least last time I checked the KFP component required to be launched into KFP)

casassg:
btw, so far unit tests work. Only change I needed to add is drop support for bytes (are harder to serialize) and move imports inside the function definition (hard one but I cant find a way to avoid this)

casassg:
So, I think this works as of now. I believe NodeWrapper should do the trick at least on KFP as it doesnt need the spec and it doesnt try to import the component class either. 

I need some ideas on how to run integration test for this though

charlesccychen:
One question here -- are we assuming the container used for execution has (the same version of) TFX installed? This use case is slightly different from the KFP one where nothing is assumed to be installed except an appropriate Python environment.

casassg:
Yes. That's the assumption. It's a bit more relaxed than: Assume the image has the module where the custom component was defined. 

Given the other components also share this assumption partially (aka the other components assume the image has X component installed) this seems hard to avoid even with other components.

casassg:
Will reopen and add more commits soon. catching up after PTO

PR #4176, opened 2021-08-21
Set parent context for pipeline and pipeline_run

casassg:
Querying pipeline runs on a pipeline is currently hard (possible but hard). But it makes sense to add a parent link between the 2. as it makes sense that a pipeline run has a parent context for the pipeline (at least logically).

I believe this code adds that logic. Which we could reuse for MLMD client on TFX-Addons


casassg:
CC @rcrowe-google could we open this one back up?

PR #4192, opened 2021-08-25
Generalize tensorboard visualization for KFP

casassg:
Currently KFP only shows Tensorboard if model has use the Trainer. This means that even when using Vertex Trainer, KFP wont show the Tensorboard. Instead of creating when Trainer component id is present, we should show the UI when model_run output is present.

We run into this with our extended Trainer since we extend Trainer by inheriting.


jiyongjung0:
This looks great! There are some remaining linting issues, I'll try to fix them.

PR #4353, opened 2021-10-07
added correctly using gpu types in vertex pipeline training notebook

entrpn:
There is an issue with tfx v1.2.0 and below where using the GPU accelerator type enum string is not recognized when running the training job in vertex ai. This is because tfx v1.2.0 and below uses a lower version of the python-aiplatform library which does not do the translation from string to enum value. It only does it for the K80 GPU which is what's displayed in this example.

So in order to use other GPU types, we must insert the actual enum value. This pr is to update the documentation so that others are aware of this work around.


entrpn:
@googlebot I signed it!

entrpn:
This should still be merged if not fixed in future tfx versions.

PR #4354, opened 2021-10-07
Change connection_config to Any message to support MLMD GRPC mode

casassg:
Currently we can't run LocalDagRunner with grpc mode and LocalDagRunner due to the proto not accepting MetadataStoreClientConfig.

Due to LocalDagRunner loading IntermediateDeploymentConfig into LocalDeploymentConfig, MLMD can't be set to use the MetadataStoreClientConfig (aka grpc mode). As seen [here](https://github.com/tensorflow/tfx/blob/08907805669f8e15764b35b2d08b54ddcae2df73/tfx/orchestration/local/runner_utils.py#L94). This is blocking our usage due to leveraging a centralized grpc server for local deployment mode for a distributed mode cache.

Would be great to have this backported to 1.2 to support usage of TF 2.5.

CC @zhitaoli @1025KB @charlesccychen


kennethyang404:
Hi Gerard, thanks for the commit! 

We already have a wrapper proto for ConnectionConfig and MetadataStoreClientConfig at [here](https://github.com/tensorflow/tfx/blob/master/tfx/proto/orchestration/execution_invocation.proto#L67-L73). Could you replace [metadata_connection_config field](https://github.com/tensorflow/tfx/blob/master/tfx/proto/orchestration/local_deployment_config.proto#L49-L50) in local_deployment_config.proto with this proto? 

Also, to do that we need to import execution_invocation.proto file into local_deployment_config.proto file, which is unnecessary. Maybe we could extract [MLMDConnectionConfig proto](https://github.com/tensorflow/tfx/blob/master/tfx/proto/orchestration/execution_invocation.proto#L67-L73) is to a separate .proto file for better code reuse in the future.

casassg:
@kennethyang404 took a stab at it. Created `metadata.proto` (not my greatest idea tbh, not sure what is a good convention for naming these). Made sure to run tests for local_dag_runner and python_binary_execution as those seemed the only touched files (but would appreciate feedback on internal CI since GitHub CI only runs on files I change).

My only worry on this (and reason I adopted previous strategy) is that this seems like a breaking change so not sure if it will affect TFX 1.X. (given its only local deployment it may not be that important since serialization is not stored anywhere afaik)

casassg:
@kennethyang404 lmk how it looks :D

casassg:
@1025KB @kennethyang404 please can this be included in the next release? Its been open for a while now...

kennethyang404 (approved):

kennethyang404 (approved):

PR #4376, opened 2021-10-14
Parametrize metadata-ui-json in KubeflowDagRunner

casassg:
Currently '/mlpipeline-ui-metadata.json' is hardcoded to be written on the root location. This may cause issues when using Kubeflow in a non-root manner.

This adds an option to parametrize the file path in the KubeflowDagRunnerConfig.


casassg:
For non-root seems we also need to mount an emptyDir to the output directory in order for argo to be able to load it back out. Implemented this in our wrapper around KubeflowDagRunner, but I would be happy to add it to the PR if wanted. Lmk!

PR #4386, opened 2021-10-19
Fixed a markdown bug

sreramk1:
Everything below the line `python penguin_pipeline_kubeflow.py --use_gcp=True` was included as "devsite-terminal", which is not the case. This also falsely gives the impression that every title below the title "Instructions for using Kubeflow V1" are subtitles or subcategories, instead of being separate titles themselves.


sreramk1:
@googlebot I signed it!

PR #4464, opened 2021-11-16
Fix error where Vertex endpoints with the same name is not deduped

jay90099:
PiperOrigin-RevId: 409465659
(cherry picked from commit 64eb496cb089221909ad50214fc2695ea7e2d213)


jiyongjung0 (approved):
Thanks!

PR #4514, opened 2021-12-07
Fix error where Vertex endpoints with the same name is not deduped

rtg0795:
PiperOrigin-RevId: 409465659


jiyongjung0 (approved):
Thanks!

Issue #4385, opened 2021-10-19
How to submit labels to AI platform training jobs?


axeltidemann:
**System information**

- Have I specified the code to reproduce the issue (Yes):
- Environment in which the code is executed (e.g., Local(Linux/MacOS/Windows),
  Interactive Notebook, Google Cloud, etc): Google Cloud
- TensorFlow version: 2.5.1
- TFX Version: 1.2.0
- Python version: 3.7.10
- Python dependencies (from `pip freeze` output):

**Describe the current behavior**

I cannot add labels to AI platform training jobs.

**Describe the expected behavior**

According to https://cloud.google.com/ai-platform/training/docs/resource-labels#add-labels, it seems like I should be able to add labels to an AI platform training job like this:

```
tfx.extensions.google_cloud_ai_platform.Trainer(...,
    custom_config = { tfx.extensions.google_cloud_ai_platform.TRAINING_ARGS_KEY: 
       { "scaleTier": ...,
         "labels": {"team": "team-name" }
       }
    }
)
```

**Standalone code to reproduce the issue**

However, the above will fail with the following error message:

```
Invalid JSON payload received. Unknown name "labels" at 'job.training_input': Cannot find field.
```

Which, according to https://cloud.google.com/ai-platform/training/docs/reference/rest/v1/projects.jobs#traininginput also makes sense. I am a bit at a loss, but I am sure it is something obvious.

axeltidemann:
[Here](https://github.com/tensorflow/tfx/blob/421501e3516fbc1b53444e7b790e61c136c90e96/tfx/extensions/google_cloud_ai_platform/runner.py#L153) it actually says: 
```
# TODO(b/168926785): Consider to change executor_class_path to job_labels.
```

axeltidemann:
The above `Labeller` custom component does not work for billing purposes, because `labels` must be set when creating the job on the AI Platform. I did not realise this until now, because there was a mixture of notebooks and jobs with the same labels that masqueraded the issue. @1025KB any thoughts on this?

axeltidemann:
I am trying to switch to `Vertex pipelines` to resolve this issue.

axeltidemann:
Vertex allows the proper use of labels. However, to use existing pipelines, I have made the necessary change (a one-liner) and created a custom TFX image for the AI Platform jobs. This will allow submission of jobs with labels to the AI Platform like so:
```
trainer_args["custom_config"] = { 
[...]

PR #4527, opened 2021-12-14
Enabling labels to be submitted to AI Platform training jobs through a TFX pipeline

axeltidemann:
This pull request is a result of a reported [issue](https://github.com/tensorflow/tfx/issues/4385) highlighting that it is not possible to submit labels with AI Platform training jobs through a TFX pipeline. This is possible on Vertex Pipelines, but a lot of pipelines will still be run on the AI Platform. The design of the solution was discussed with @1025KB, and we agreed on a small change that allows the `Trainer` component to be called with `labels` like this: 

```
tfx.extensions.google_cloud_ai_platform.Trainer(...,
    custom_config = { tfx.extensions.google_cloud_ai_platform.TRAINING_ARGS_KEY: { "scaleTier": ...},
                      "labels": {"team": "awesome" }
    }
)
```

Labels are very useful for billing purposes, which is why this change is desirable.

### DIFF
--- a/tfx/extensions/google_cloud_ai_platform/runner.py
+++ b/tfx/extensions/google_cloud_ai_platform/runner.py
@@ -156,8 +156,8 @@ def start_cloud_training(input_dict: Dict[str, List[types.Artifact]],
                          exec_properties: Dict[str, Any],
                          executor_class_path: str,
                          job_args: Dict[str, Any],
-                         job_labels: Dict[str, Any],
                          job_id: Optional[str],
+                         job_labels: Optional[Dict[str, Any]] = None, # AI Platform only.
                          enable_vertex: Optional[bool] = False,
                          vertex_region: Optional[str] = None):
   """Start a trainer job on AI Platform (AIP).
@@ -177,12 +177,12 @@ def start_cloud_training(input_dict: Dict[str, List[types.Artifact]],
          For the full set of parameters supported by Google Cloud AI Platform
          (CAIP) TrainingInput, refer to
        https://cloud.google.com/ml-engine/docs/tensorflow/training-jobs#configuring_the_job
-    job_labels: Labels for AI Platform training job.
     job_id: Job ID for AI Platform Training job. If not supplied,
       system-determined unique ID is given. Refer to
       https://cloud.google.com/ml-engine/reference/rest/v1/projects.jobs#resource-job.
         In Vertex AI, the job_id corresponds to the display name, a unique ID is
         always given to the created job.
+    job_labels: Labels for AI Platform training job.
     enable_vertex: Whether to enable Vertex or not.
     vertex_region: Region for endpoint in Vertex training.
 
@@ -196,7 +196,7 @@ def start_cloud_training(input_dict: Dict[str, List[types.Artifact]],
   training_job = client.create_training_job(input_dict, output_dict,
                                             exec_properties,
                                             executor_class_path, job_args,
-                                            job_labels, job_id)
+                                            job_id, job_labels)
 
   _launch_cloud_training(
       project=project,
--- a/tfx/extensions/google_cloud_ai_platform/trainer/executor.py
+++ b/tfx/extensions/google_cloud_ai_platform/trainer/executor.py
@@ -119,8 +119,8 @@ class GenericExecutor(base_executor.BaseExecutor):
                                      executor_class.__name__)
     # Note: exec_properties['custom_config'] here is a dict.
     return runner.start_cloud_training(input_dict, output_dict, exec_properties,
-                                       executor_class_path, job_args, job_labels, 
-                                       job_id, enable_vertex,
+                                       executor_class_path, job_args, 
+                                       job_id, job_labels, enable_vertex,
                                        vertex_region)
 
 
--- a/tfx/extensions/google_cloud_ai_platform/training_clients.py
+++ b/tfx/extensions/google_cloud_ai_platform/training_clients.py
@@ -70,8 +70,8 @@ class AbstractJobClient(abc.ABC):
 
   @abc.abstractmethod
   def create_training_job(self, input_dict, output_dict, exec_properties,
-                          executor_class_path, job_args, job_labels,
-                          job_id) -> Dict[str, Any]:
+                          executor_class_path, job_args, 
+                          job_id, job_labels) -> Dict[str, Any]:
     """Get training args for runner._launch_aip_training.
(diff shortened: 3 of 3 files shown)


axeltidemann:
I have submitted my Contributor License Agreement. Please re-scan.

tfx/extensions/google_cloud_ai_platform/training_clients.py
```diff
@@ -228,6 +230,10 @@ def create_training_job(self, input_dict: Dict[str, List[types.Artifact]],
         {telemetry_utils.LABEL_TFX_EXECUTOR: executor_class_path}):
       job_labels = telemetry_utils.make_labels_dict()
 
+    custom_config = json_utils.loads(
+        exec_properties.get(standard_component_specs.CUSTOM_CONFIG_KEY))
+    job_labels.update(custom_config.get('labels', {}))
```
1025KB, line 235: for cutom_config level keys, you would want to add it like [this](https://github.com/tensorflow/tfx/blob/master/tfx/extensions/google_cloud_ai_platform/trainer/executor.py#L40), thus users can find the keys and use it. consider make it an optional param to pass in this function from executor.py

axeltidemann: Good point @1025KB, I have added the `LABELS_KEY` to accommodate this. Minor logic change to merge labels in `executor.py`. In the `CAIPJobClient` I simply pop the `labels` dictionary from the `training_inputs` and update the `job_labels` dictionary.

tfx/extensions/google_cloud_ai_platform/trainer/executor.py
```diff
@@ -97,6 +104,9 @@ def Do(self, input_dict: Dict[str, List[types.Artifact]],
       absl.logging.error(err_msg)
       raise ValueError(err_msg)
 
+    job_args['labels'] = { **job_args.get('labels', {}),
+                           **custom_config.get(LABELS_KEY, {}) }
```
1025KB, line 108: Let's pass it in just like job_id (line 110)? or do it like [project](https://github.com/tensorflow/tfx/blob/master/tfx/extensions/google_cloud_ai_platform/runner.py#L191)

as the [job_args](https://github.com/tensorflow/tfx/blob/f84b768f28ed60ab1b51dded73c64718950db1f4/tfx/extensions/google_cloud_ai_platform/training_clients.py#L200) of client.create_training_job is the training_input, so I think it would be good to keep it as it is, and add an optional labels to that function?

axeltidemann: 1. If it is passed in like `job_id`, then the signatures of `runner.start_cloud_training`, `CAIPJobClient.create_training_job` and `VertexJobClient.create_training_job` would have to change, and it seems a bit convoluted, since `labels` is part of `job_args` for `VertexJobClient` already?
2. Similar argument for doing it like `project`, but it would be slightly simpler than option 1 since you would not have to change `runner.start_cloud_training` in addition.

I see your point about the `job_args` in `CAIPJobClient` should not have `labels`, but it is simply stored there and [popped](https://github.com/axeltidemann/tfx/blob/18f05ff652919f3e8566afd583e4fb860e29b392/tfx/extensions/google_cloud_ai_platform/training_clients.py#L230) before passed on - which is similar to the "stowaway" project you refer to. I personally think my proposal is the least "intrusive" option, but I have implemented option 1 for you to review.

1025KB: I see your point. if we want to just pass the label in job_args as a workaround for CAIP, then we don't need to add constant to executor.py, those constants are top level keys of custom_config

axeltidemann: So you mean like my first commit, basically?

axeltidemann: I will go on holiday leave now, and will check back in two weeks. If I understand you correctly, what you are describing is my first commit - feel free to approve that in the meantime. If not, I am of course happy to continue this when I am back. Happy holidays!

1025KB: Current code look good (except we want to make labels an optional args for AI platform only)

(Due to Vertex/AI platform diff, I don't think we have a perfect way to align them)

tfx/extensions/google_cloud_ai_platform/runner.py
```diff
@@ -156,6 +156,7 @@ def start_cloud_training(input_dict: Dict[str, List[types.Artifact]],
                          exec_properties: Dict[str, Any],
                          executor_class_path: str,
                          job_args: Dict[str, Any],
+                         job_labels: Dict[str, Any],
```
1025KB, line 159: can we make job_lables optional (same elsewhere), and comment it's for AI platform only? Thanks!

axeltidemann:
I am working on this, will update as soon as I can.

tfx/extensions/google_cloud_ai_platform/trainer/executor.py
axeltidemann: I have made `job_labels` optional, please review.

axeltidemann:
@1025KB Could you review the changes I made, please?

axeltidemann:
@1025KB Please have a look - it would be awesome to finish this.

1025KB (approved):

tfx/extensions/google_cloud_ai_platform/runner.py
```diff
@@ -157,6 +157,7 @@ def start_cloud_training(input_dict: Dict[str, List[types.Artifact]],
                          executor_class_path: str,
                          job_args: Dict[str, Any],
                          job_id: Optional[str],
+                         job_labels: Optional[Dict[str, Any]] = None, # AI Platform only.
```
chensun, line 160: Adding a new parameter in the middle would be a breaking change for those rely on positional arguments. Do we have any concern on this? 
/cc @1025KB

axeltidemann: I added it there because it semantically would make more sense to have `job_args, job_id, job_labels` grouped together. It can be moved to the end, of course. Or could this be listed in the `RELEASE.md` under `Breaking Changes`?

chensun: Sounds good then. Thanks!

PR #4528, opened 2021-12-14
Add exit handler to kf_dag_runner'

casassg:
Adds exit handler to KFv1 DagRunner. This should keep parity with KFv2 exit handler as it's pretty similar. 

Still needs a way to pass the end status string probably (will need to modify component at compile time most likely).

Tested: `python tfx/orchestration/kubeflow/kubeflow_dag_runner_test.py`

### DIFF
--- a/tfx/orchestration/kubeflow/base_component.py
+++ b/tfx/orchestration/kubeflow/base_component.py
@@ -30,6 +30,7 @@ from kubernetes import client as k8s_client
 from tfx.dsl.components.base import base_node as tfx_base_node
 from tfx.orchestration import data_types
 from tfx.orchestration import pipeline as tfx_pipeline
+from tfx.orchestration.kubeflow import utils
 from tfx.orchestration.kubeflow.proto import kubeflow_pb2
 from tfx.proto.orchestration import pipeline_pb2
 
@@ -92,6 +93,8 @@ class BaseComponent:
       runtime_parameters: Runtime parameters of the pipeline.
       metadata_ui_path: File location for metadata-ui-metadata.json file.
     """
+
+    utils.replace_placeholder(component)
 
     arguments = [
         '--pipeline_root',
--- a/tfx/orchestration/kubeflow/base_component.py
+++ b/tfx/orchestration/kubeflow/base_component.py
@@ -30,7 +30,6 @@ from kubernetes import client as k8s_client
 from tfx.dsl.components.base import base_node as tfx_base_node
 from tfx.orchestration import data_types
 from tfx.orchestration import pipeline as tfx_pipeline
-from tfx.orchestration.kubeflow import utils
 from tfx.orchestration.kubeflow.proto import kubeflow_pb2
 from tfx.proto.orchestration import pipeline_pb2
 
@@ -93,8 +92,6 @@ class BaseComponent:
       runtime_parameters: Runtime parameters of the pipeline.
       metadata_ui_path: File location for metadata-ui-metadata.json file.
     """
-
-    utils.replace_placeholder(component)
 
     arguments = [
         '--pipeline_root',
--- a/tfx/orchestration/kubeflow/kubeflow_dag_runner.py
+++ b/tfx/orchestration/kubeflow/kubeflow_dag_runner.py
@@ -34,6 +34,7 @@ from tfx.orchestration.config import pipeline_config
 from tfx.orchestration.kubeflow import base_component
 from tfx.orchestration.kubeflow import utils
 from tfx.orchestration.kubeflow.proto import kubeflow_pb2
+from tfx.orchestration.kubeflow.v2.compiler_utils import TFX_DAG_NAME
 from tfx.orchestration.launcher import base_component_launcher
 from tfx.orchestration.launcher import in_process_component_launcher
 from tfx.orchestration.launcher import kubernetes_component_launcher
@@ -358,7 +359,7 @@ class KubeflowDagRunner(tfx_runner.TfxRunner):
     if self._exit_handler:
       exit_op = component_to_kfp_op[self._exit_handler]
       with dsl.ExitHandler(exit_op) as exit_handler_group:
-        exit_handler_group.name = '_tfx_dag'
+        exit_handler_group.name = TFX_DAG_NAME
         # KFP get_default_pipeline should have the pipeline object when invoked while compiling
         # This allows us to retrieve all ops from pipeline group (should be the only group in the pipeline)
         pipeline_group = dsl.Pipeline.get_default_pipeline().groups[0]
--- a/tfx/orchestration/kubeflow/kubeflow_dag_runner.py
+++ b/tfx/orchestration/kubeflow/kubeflow_dag_runner.py
@@ -32,6 +32,7 @@ from tfx.orchestration import pipeline as tfx_pipeline
 from tfx.orchestration import tfx_runner
 from tfx.orchestration.config import pipeline_config
 from tfx.orchestration.kubeflow import base_component
+from tfx.orchestration.kubeflow import utils
 from tfx.orchestration.kubeflow.proto import kubeflow_pb2
 from tfx.orchestration.launcher import base_component_launcher
 from tfx.orchestration.launcher import in_process_component_launcher
@@ -317,6 +318,9 @@ class KubeflowDagRunner(tfx_runner.TfxRunner):
       pipeline_root: dsl.PipelineParam representing the pipeline root.
     """
     component_to_kfp_op = {}
+    
+    for component in pipeline.components:
+      utils.replace_placeholder(component)
     tfx_ir = self._generate_tfx_ir(pipeline)
 
     # Assumption: There is a partial ordering of components in the list, i.e.,
--- a/tfx/orchestration/kubeflow/kubeflow_dag_runner.py
+++ b/tfx/orchestration/kubeflow/kubeflow_dag_runner.py
@@ -34,7 +34,7 @@ from tfx.orchestration.config import pipeline_config
 from tfx.orchestration.kubeflow import base_component
 from tfx.orchestration.kubeflow import utils
 from tfx.orchestration.kubeflow.proto import kubeflow_pb2
-from tfx.orchestration.kubeflow.v2.compiler_utils import TFX_DAG_NAME
+from tfx.orchestration.kubeflow.utils import TFX_DAG_NAME
 from tfx.orchestration.launcher import base_component_launcher
 from tfx.orchestration.launcher import in_process_component_launcher
 from tfx.orchestration.launcher import kubernetes_component_launcher
--- a/tfx/orchestration/kubeflow/kubeflow_dag_runner.py
+++ b/tfx/orchestration/kubeflow/kubeflow_dag_runner.py
@@ -410,7 +410,7 @@ class KubeflowDagRunner(tfx_runner.TfxRunner):
     if self._exit_handler:
       original_pipeline = pipeline
       pipeline = copy.copy(original_pipeline)
-      pipeline.components = [self._exit_handler] + original_pipeline.components
+      pipeline.components.append(self._exit_handler)
 
     for component in pipeline.components:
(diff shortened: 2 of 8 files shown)


1025KB:
Thanks!!

tfx/orchestration/kubeflow/kubeflow_dag_runner.py
```diff
@@ -347,6 +350,13 @@ def _construct_pipeline_graph(self, pipeline: tfx_pipeline.Pipeline,
 
       component_to_kfp_op[component] = kfp_component.container_op
 
+    # If exit handler defined create an exit handler and add all ops to it.
+    if self._exit_handler:
+      exit_op = component_to_kfp_op[self._exit_handler]
```
1025KB, line 355: I might miss it, where does the exit_handler being converted to kfp ops?

How does FinalStatusStr get passed in to exit handler?

casassg: It's being transformed as part of the main for loop that creates BaseComponent (since I added the exit_handler in the pipeline).

FinalStatusStr is what I need to figure out still. I think I can make it work but I need to modify a exec property value on the fly.

casassg: Found how to replace FinalStatusStr. I used tfx/orchestration/kubeflow/utils.py to fix this. I basically just replace the class for a string that Argo will replace w status: `'{{workflow.status}}'`

See: https://github.com/argoproj/argo-workflows/blob/master/docs/variables.md#exit-handler for argo. 
Tekton has a different one: https://github.com/tektoncd/pipeline/blob/main/docs/variables.md#variables-available-in-a-pipeline

tfx/orchestration/kubeflow/base_component.py
```diff
@@ -94,8 +93,6 @@ def __init__(self,
       metadata_ui_path: File location for metadata-ui-metadata.json file.
     """
 
-    utils.replace_placeholder(component)
```
casassg, line 97: This was a noop given that it was changing component.exec_properties and it's not being used before. 

This caused me to not understand the failure mode, so removing it to make it clear.

tfx/orchestration/kubeflow/utils.py
```diff
@@ -20,17 +20,16 @@
...
+  """Replaces the TFX placeholders with KFP placeholders"""
   keys = list(component.exec_properties.keys())
   for key in keys:
     exec_property = component.exec_properties[key]
-    if not isinstance(exec_property, data_types.RuntimeParameter):
-      continue
-    component.exec_properties[key] = str(
-        dsl.PipelineParam(name=exec_property.name))
+    if isinstance(exec_property, FinalStatusStr):
+      component.exec_properties[key] = '{{workflow.status}}'
```
casassg, line 32: This function was only being used in base_component but it was not having any effect there, so moved it up and used it for FinalStatusStr

tfx/orchestration/kubeflow/kubeflow_dag_runner.py
```diff
@@ -347,6 +350,13 @@ def _construct_pipeline_graph(self, pipeline: tfx_pipeline.Pipeline,
...
+      with dsl.ExitHandler(exit_op) as exit_handler_group:
+        exit_handler_group.name = 'tfx-exit-handler'
```
polydier1, line 361: for consistency with V2, we may call it '_tfx_dag'
(note:variable for v2 is defined under orchestration/kubeflow/v2/compiler_utils.py)

tfx/orchestration/kubeflow/kubeflow_dag_runner.py
polydier1: Thanks for the pull.
The container_op converting seems right to me.

Has the logic about FinalStatusStr reflected in the pull yet? I didn't spot that.

It would be helpful to add the corresponding test with exit handler in /tfx/orchestration/kubeflow/kubeflow_dag_runner_test.py to validate the correctness.

1025KB: Thanks, sorry didn't notice the component is added to pipeline at line 413

Do you have a pic of the execution graph of an e2e pipeline with exit handler in kfp? we want to see the relation between node and see if it align with Vertex.

casassg: @polydier1 yes, it has been updated. See utils.py to see where the logic for FinalStatusStr. Also added a test here (which passes).

I have a pic of a sample pipeline I was using w similar code internally (It may have different names since its not an exact copy from here, and its failed due to me messing up the docker image).

Screen Shot 2021-12-16 at 8 20 41 PM


Screen Shot 2021-12-16 at 8 20 30 PM

1025KB: Interesting, in Vertex, the pipeline itself is treated as a whole, and exit handler will be executed even if pipeline failed to execute to the end.

Seems in this graph, exit_handler is treated as the downstream of filepusher? what if trainer throws an error and pusher doesn't get executed?

casassg: it does get executed still. For some reason the UI shows it like this (see the other graph)

tfx/orchestration/kubeflow/utils.py
1025KB: This is for supporting runtime param if I remember correctly, could you create a separate function instead of changing this?

casassg: I believe this function in the current codebase is a noop, you can see how below in the component those are not used. That said, I can derisk this by making it a separate function

tfx/orchestration/kubeflow/kubeflow_dag_runner.py
1025KB: Thank!!! I think this looks good in general!

(Will let ExitHandler owner take a look when she's back from vacation)

casassg: sounds good! happy holidays y'all!

casassg:
Just realized, exit handler should never have cache since we want it to execute always. 

This seems to be an error on KFP v2 as well btw

1025KB:
> Just realized, exit handler should never have cache since we want it to execute always.
> 
> This seems to be an error on KFP v2 as well btw

Nice found! what you can do is make sure (assert) the generated IR set enable_cache=false for your exit_handler component

casassg:
yup. Added on the kubeflow_dag_runner_test.

That said, if im reading correctly kfp v2 may have a similar issue while it may be easier to fix though

1025KB:
> yup. Added on the kubeflow_dag_runner_test.
> 
> That said, if im reading correctly kfp v2 may have a similar issue while it may be easier to fix though

Yep, [here](https://github.com/tensorflow/tfx/blob/master/tfx/orchestration/kubeflow/v2/step_builder.py#L364) I think , will check with owner after holiday. Thanks!!

casassg:
Actually, it may be simpler: https://github.com/tensorflow/tfx/blob/master/tfx/orchestration/kubeflow/v2/pipeline_builder.py#L163

gbaned:
@zhitaoli Can you please review this PR ? Thanks!

tfx/orchestration/kubeflow/kubeflow_dag_runner.py
```diff
@@ -25,12 +26,15 @@
...
 from tfx.dsl.components.base import base_component as tfx_base_component
+from tfx.dsl.components.base import base_node
 from tfx.orchestration import data_types
 from tfx.orchestration import pipeline as tfx_pipeline
 from tfx.orchestration import tfx_runner
 from tfx.orchestration.config import pipeline_config
 from tfx.orchestration.kubeflow import base_component
+from tfx.orchestration.kubeflow import utils
 from tfx.orchestration.kubeflow.proto import kubeflow_pb2
+from tfx.orchestration.kubeflow.v2.compiler_utils import TFX_DAG_NAME
```
polydier1, line 37: this name is better defined in the common file (eg.  tfx/orchestration/kubeflow/utils.py) instead imported from v2 folder.

If you don't want to include change in current pull, I can refactor later. 
Just heads up here.

casassg: That's fair. I'm happy to include it. I was a bit wary of adding too many changes for this

tfx/orchestration/kubeflow/kubeflow_dag_runner_test.py
```diff
@@ -36,6 +39,9 @@
 
 from ml_metadata.proto import metadata_store_pb2
 
+@component
```
polydier1, line 42: shall we change the decorator as @exit_handler(imported from tfx/v1/orchestration/experimental/__init__.py)

tfx/orchestration/kubeflow/kubeflow_dag_runner_test.py
```diff
@@ -290,5 +296,32 @@ def testContainerComponent(self):
...
+          c for c in pipeline['spec']['templates'] if 'container' in c
+      ]
+      self.assertLen(containers, 2)
+      exit_component_args = ' '.join(containers[1]['container']['args'])
+      self.assertIn('{{workflow.status}}', exit_component_args)
+      self.assertNotIn('enableCache', exit_component_args)
+      first_component_args = ' '.join(containers[0]['container']['args'])
+      self.assertNotIn('{{workflow.status}}', first_component_args)
+      self.assertIn('enableCache', first_component_args)
+
```
polydier1, line 324: Thanks for adding tests:)

tfx/orchestration/kubeflow/kubeflow_dag_runner.py
polydier1: let's double check the graph is correct before merging it. I have forwarded the picture to KFP team to verify.

polydier1 (commented):
Generally looks good to me, left some minor comments. 
For UI, I am verifying with KFP team to make sure it's correct.

polydier1:
> Actually, it may be simpler: https://github.com/tensorflow/tfx/blob/master/tfx/orchestration/kubeflow/v2/pipeline_builder.py#L163

Thanks for detecting that, I will make the change for V2 to not cache.

tfx/orchestration/kubeflow/kubeflow_dag_runner.py
polydier1: you can define same name and import it on compile_utils, I can make change imports of v2 to avoid making the pull too big.

tfx/orchestration/kubeflow/kubeflow_dag_runner_test.py
casassg: I was mostly following the [test in v2](https://github.com/tensorflow/tfx/blob/3288393efc889f9db171d3fd81f9dc818a9fce8a/tfx/orchestration/kubeflow/v2/test_utils.py#L326) which uses a container component. Not opinionated here so I can change if preferred.

polydier1: I think @exit_handler is preferred here, it will perform exactly as @component, it's just a recommended syntax for exit handler component.

polydier1:
Thanks for the update. The pull looks good to me, so we are waiting:
1. KFP team to confirm the UI is correct, will follow up after the KFP owner come back from holiday vacation
2. after 1 is confirmed, I will need to get authorization to merge the pull request, also need to wait for repo owner come back from holiday vacation.

polydier1 (approved):
Thanks!
Double checked the Kubeflow team, the graph is correct, there is no DAG UI in KFP currently.

casassg:
any update on merging this before 1.6?

casassg:
👀

1025KB (approved):

1025KB:
ready to merge

casassg:
Re-up

jay90099 (approved):

Issue #3732, opened 2021-05-13
Support dumping custom Kubeflow ui metadata


jeongukjae:
**System information**

- TFX Version (you are using): 0.29
- Environment in which you plan to use the feature (e.g., Local
  (Linux/MacOS/Windows), Interactive Notebook, Google Cloud, etc..): Google Cloud
- Are you willing to contribute it (Yes/No): Yes

**Describe the feature and the current behavior/state.**

https://github.com/tensorflow/tfx/blob/79fa0d8adce177d04feaa737f1440689344521df/tfx/orchestration/kubeflow/container_entrypoint.py#L210-L296

For now, `_dump_ui_metadata` function in `tfx/orchestration/kubeflow/container_entrypoint.py` does not consider already existing `/mlpipeline-ui-metadata.json`. So if I write some metadata that I want to see in KubeFlow UI in a custom component, the file will be overwritten.

If `/mlpipeline-ui-metadata.json` already exists(maybe created in the component) and has a valid format, I think extending an existing file is better.

**Will this change the current API? How?**

No

**Who will benefit with this feature?**

**Do you have a workaround or are completely blocked by this?** :

**Name of your Organization (Optional)**

**Any Other info.**

jeongukjae:
@kennethyang404 is there any progress here?

jiyongjung0:
Sorry for the late response. Your suggestion sounds great. If you can provide some PR, I'm happy to review and merge it. 

cc. @chensun @zijianjoy @Bobgy

zijianjoy:
Sounds good to me for extending `mlpipeline_ui_metadata.json` as it supports multiple visualizations in `outputs`.

PR #4674, opened 2022-02-10
Update `_dump_ui_metadata` function to extend existing UI metadata file

jeongukjae:
In the current implementation, `_dump_ui_metadata` function in the KubeFlow container entrypoint does not consider the already existing UI metadata file. Therefore, we cannot use custom UI metadata in our own components.

So I submit this PR patches to update the function to extend existing UI metadata items.

Resolve https://github.com/tensorflow/tfx/issues/3732

### DIFF
--- a/tfx/orchestration/kubeflow/container_entrypoint.py
+++ b/tfx/orchestration/kubeflow/container_entrypoint.py
@@ -243,6 +243,10 @@ def _dump_ui_metadata(
   For general components we just render a simple Markdown file for
     exec_properties/inputs/outputs.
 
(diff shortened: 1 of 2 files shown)


casassg:
@rcrowe-google This would be great to consider merging (I've personally have given this one a thought but never was able to figure out what was the best way to do it)

jiyongjung0 (approved):
LGTM!

Thank you for the contribution. I might change a little bit before actually merging this.
The automated check is failing for dependency issues, and I'll work on that, too.

PR #4761, opened 2022-03-24
Extend tfxio factory to use parquet-tfxio

martinbomio:
This PR depends on https://github.com/tensorflow/tfx-bsl/pull/53 and https://github.com/tensorflow/tfx-bsl/pull/54 being released.

Extend the `make_tfxio` factory to use `ParquetTFXIO` if the `payload_format` is `FORMAT_PARQUET`.

### DIFF
--- a/tfx/components/util/tfxio_utils.py
+++ b/tfx/components/util/tfxio_utils.py
@@ -23,6 +23,7 @@ from tfx.proto import example_gen_pb2
 from tfx.types import artifact
 from tfx.types import standard_artifacts
 from tfx_bsl.tfxio import dataset_options
+from tfx_bsl.tfxio import parquet_tfxio
 from tfx_bsl.tfxio import raw_tf_record
 from tfx_bsl.tfxio import record_to_tensor_tfxio
 from tfx_bsl.tfxio import tf_example_record
@@ -30,6 +31,8 @@ from tfx_bsl.tfxio import tf_sequence_example_record
 from tfx_bsl.tfxio import tfxio
 from tensorflow_metadata.proto.v0 import schema_pb2
 
+
+_SUPPORTED_PAYLOAD_FORMATS = ['parquet', 'tfrecords_gzip']
 # TODO(b/162532479): switch to support List[str] exclusively, once tfx-bsl
 # post-0.22 is released.
 OneOrMorePatterns = Union[str, List[str]]
@@ -272,7 +275,7 @@ def make_tfxio(
       option, and an error will be raised in that case. Required if
       read_as_raw_records == True.
     file_format: file format string for each file_pattern. Only 'tfrecords_gzip'
-      is supported for now.
+      and 'parquet' are supported for now.
 
   Returns:
     a TFXIO instance.
@@ -291,10 +294,10 @@ def make_tfxio(
             f'The length of file_pattern and file_formats should be the same.'
             f'Given: file_pattern={file_pattern}, file_format={file_format}')
       else:
-        if any(item != 'tfrecords_gzip' for item in file_format):
+        if any(item in _SUPPORTED_PAYLOAD_FORMATS for item in file_format):
           raise NotImplementedError(f'{file_format} is not supported yet.')
     else:  # file_format is str type.
-      if file_format != 'tfrecords_gzip':
+      if file_format in _SUPPORTED_PAYLOAD_FORMATS:
         raise NotImplementedError(f'{file_format} is not supported yet.')
 
   if read_as_raw_records:
@@ -329,6 +332,12 @@ def make_tfxio(
         saved_decoder_path=data_view_uri,
         telemetry_descriptors=telemetry_descriptors,
         raw_record_column_name=raw_record_column_name)
+
+  if payload_format == example_gen_pb2.PayloadFormat.FORMAT_PARQUET:
+    return parquet_tfxio.ParquetTFXIO(
+      file_pattern=file_pattern,
+      schema=schema,
+      telemetry_descriptors=telemetry_descriptors)
 
   raise NotImplementedError(
       'Unsupport payload format: {}'.format(payload_format))
--- a/tfx/components/util/tfxio_utils.py
+++ b/tfx/components/util/tfxio_utils.py
@@ -32,7 +32,7 @@ from tfx_bsl.tfxio import tfxio
 from tensorflow_metadata.proto.v0 import schema_pb2
 
 
-_SUPPORTED_PAYLOAD_FORMATS = ['parquet', 'tfrecords_gzip']
+_SUPPORTED_FILE_FORMATS = {example_gen_pb2.FileFormat.FORMAT_PARQUET, example_gen_pb2.FileFormat.FORMAT_TFRECORDS_GZIP}
 # TODO(b/162532479): switch to support List[str] exclusively, once tfx-bsl
 # post-0.22 is released.
 OneOrMorePatterns = Union[str, List[str]]
@@ -245,12 +245,12 @@ def get_data_view_decode_fn_from_artifact(
 def make_tfxio(
     file_pattern: OneOrMorePatterns,
     telemetry_descriptors: List[str],
-    payload_format: Union[str, int],
+    payload_format: Union[example_gen_pb2.PayloadFormat, int],
     data_view_uri: Optional[str] = None,
     schema: Optional[schema_pb2.Schema] = None,
     read_as_raw_records: bool = False,
     raw_record_column_name: Optional[str] = None,
-    file_format: Optional[Union[str, List[str]]] = None) -> tfxio.TFXIO:
+    file_format: Optional[Union[example_gen_pb2.FileFormat, List[example_gen_pb2.FileFormat]]] = None) -> tfxio.TFXIO:
   """Creates a TFXIO instance that reads `file_pattern`.
 
   Args:
@@ -274,7 +274,7 @@ def make_tfxio(
       that column will be the raw records. Note that not all TFXIO supports this
       option, and an error will be raised in that case. Required if
       read_as_raw_records == True.
-    file_format: file format string for each file_pattern. Only 'tfrecords_gzip'
+    file_format: file format for each file_pattern. Only 'tfrecords_gzip'
       and 'parquet' are supported for now.
 
   Returns:
@@ -294,10 +294,10 @@ def make_tfxio(
             f'The length of file_pattern and file_formats should be the same.'
             f'Given: file_pattern={file_pattern}, file_format={file_format}')
       else:
-        if any(item in _SUPPORTED_PAYLOAD_FORMATS for item in file_format):
+        if any(item in _SUPPORTED_FILE_FORMATS for item in file_format):
           raise NotImplementedError(f'{file_format} is not supported yet.')
     else:  # file_format is str type.
-      if file_format in _SUPPORTED_PAYLOAD_FORMATS:
+      if file_format in _SUPPORTED_FILE_FORMATS:
         raise NotImplementedError(f'{file_format} is not supported yet.')
 
   if read_as_raw_records:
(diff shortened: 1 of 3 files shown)


martinbomio:
@iindyk PTAL

tfx/components/util/tfxio_utils.py
```diff
@@ -23,13 +23,16 @@
...
+from tfx_bsl.tfxio import parquet_tfxio
 from tfx_bsl.tfxio import raw_tf_record
 from tfx_bsl.tfxio import record_to_tensor_tfxio
 from tfx_bsl.tfxio import tf_example_record
 from tfx_bsl.tfxio import tf_sequence_example_record
 from tfx_bsl.tfxio import tfxio
 from tensorflow_metadata.proto.v0 import schema_pb2
 
+
+_SUPPORTED_PAYLOAD_FORMATS = ['parquet', 'tfrecords_gzip']
```
iindyk, line 35: nit: set or tuple

1025KB: consider to use [this](https://github.com/tensorflow/tfx/blob/master/tfx/proto/example_gen.proto#L123)

FORMAT_TFRECORDS_GZIP is currently supported, and FORMAT_PARQUET will be added.
ExampleGen will attach the file format in its output example artifact, which can be used by downstream to figure out the example artifact's file format

tfx/proto/example_gen.proto
```diff
@@ -111,7 +111,10 @@ enum PayloadFormat {
   // Serialized any protocol buffer.
   FORMAT_PROTO = 11;
 
-  reserved 1 to 5, 8 to 10, 12 to max;
+  // Serialized parquet messages.
+  FORMAT_PARQUET = 12;
```
1025KB, line 115: File format is also needed?

e.g., FORMAT_TF_EXAMPLE (payload format) is stored in FORMAT_TFRECORDS_GZIP (file format)

for Parquet, what's the file format?

tfx/components/util/tfxio_utils.py
martinbomio: ah nice, I missed that FileFormat proto message!

tfx/proto/example_gen.proto
martinbomio: I guess PARQUET is also the file format?

martinbomio: @1025KB these are the file formats supported by pyarrow https://arrow.apache.org/docs/python/parquet.html#compression-encoding-and-file-compatibility. Should I add one for each? or should we agree on a set of ones we are planning to support?

martinbomio: the default is SNAPPY

1025KB: Not necessary, one splittable file format (parquet) in addition to tfrecord (non-splittable file format) should be enough for now

1025KB: Note that we will use beam.io.writeToParquet to generate output files, so for reading, only need to support the format used by beam.io.writeToParquet

rcrowe-google: While we're in this code, to maximize compatibility I think it would make sense to include gzip, brotli, and none, as well as snappy - unless that creates headaches, in which case snappy and none (and maybe gzip?) would seem sufficient.

tfx/components/util/tfxio_utils.py
martinbomio: @1025KB wondering if you also wanted me to change the type of `file_format` param in `make_tfxio` to `example_gen_pb2.FileFormat`?

1025KB: oh, I didn't notice it's Payload format, then using [this](https://github.com/tensorflow/tfx/blob/master/tfx/proto/example_gen.proto#L101) payload format is fine too

what I mean here is instead of hardcoded 'parquet', 'tfrecords_gzip' str, can we use enum in PayloadFormat proto here? otherwise I feel there will be a PayloadFormat.FORMAT_PARQUET to 'parquet' conversion somewhere?

iindyk@, any thoughts?

martinbomio: done in https://github.com/tensorflow/tfx/pull/4761/commits/465ddff65590050c4a1adb74aeb7b01336af539c. The variable was actually badly named, it is indeed the supported file formats

iindyk: In general I think it makes sense to have enum for such thing instead of string, but if we want to update this we'd also need to update all callsites: components and tests, which is a bit trickier. If this proves to be messy (I think it may be) I'd keep it as is for now and add a todo to refactor this.

1025KB: I see, then a todo sound good

martinbomio: @iindyk should i revert the last commit and add a TODO then?

iindyk: > should i revert the last commit and add a TODO then?
sgtm

> I did check the calls to make_tfxio within this project and non of them were passing the file format

Even tests? that's unfortunate this path is not covered. Aside from the upstream code we'd also need to update tfxios that are produced by the factory, bc they take strings (or make a bridge proto enum ->string).

For payload_format there are places that actually pass int (and I'm not yet sure if there's a good reason for that, this seems fragile).

martinbomio:
@iindyk any update on the release?

iindyk:
The release is currently blocked on a tf-related breakage, but will likely happen this week

martinbomio:
@iindyk any update on the release?

iindyk:
I think all the release blockers are resolved at this point, so our release engineers are working on it, should be coming out soon

martinbomio:
hey @iindyk should we think about wrapping this one up? Are there any outstanding comments in the PR that we need to address to be able to merge it?

tfx/components/util/tfxio_utils.py
```diff
@@ -23,13 +23,16 @@
...
+from tfx_bsl.tfxio import parquet_tfxio
 from tfx_bsl.tfxio import raw_tf_record
 from tfx_bsl.tfxio import record_to_tensor_tfxio
 from tfx_bsl.tfxio import tf_example_record
 from tfx_bsl.tfxio import tf_sequence_example_record
 from tfx_bsl.tfxio import tfxio
 from tensorflow_metadata.proto.v0 import schema_pb2
 
+
+_SUPPORTED_FILE_FORMATS = {example_gen_pb2.FileFormat.FORMAT_PARQUET, example_gen_pb2.FileFormat.FORMAT_TFRECORDS_GZIP}
```
iindyk, line 35: nit: use immutable tuple for constants

tfx/proto/example_gen.proto
iindyk: I think supporting only the format that beam source and sink support should be sufficient for now, other formats could be considered in a follow up if needed

tfx/components/util/tfxio_utils.py
```diff
@@ -291,10 +294,10 @@ def make_tfxio(
             f'The length of file_pattern and file_formats should be the same.'
             f'Given: file_pattern={file_pattern}, file_format={file_format}')
       else:
-        if any(item != 'tfrecords_gzip' for item in file_format):
+        if any(item in _SUPPORTED_FILE_FORMATS for item in file_format):
```
iindyk, line 297: not in?

tfx/components/util/tfxio_utils.py
```diff
@@ -291,10 +294,10 @@ def make_tfxio(
...
           raise NotImplementedError(f'{file_format} is not supported yet.')
     else:  # file_format is str type.
-      if file_format != 'tfrecords_gzip':
+      if file_format in _SUPPORTED_FILE_FORMATS:
```
iindyk, line 300: ditto

iindyk (commented):
left couple comments, but generally LGTM. I don't have owner permission here, so @1025KB PTAL

tfx/proto/example_gen.proto
```diff
@@ -111,7 +111,10 @@ enum PayloadFormat {
...
+
+  reserved 1 to 5, 8 to 10, 13 to max;
```
1025KB, line 117: Let's use 15

reserved 1 to 5, 8 to 10, 12 to 14, 16 to max;

tfx/proto/example_gen.proto
```diff
@@ -122,7 +125,11 @@ enum FileFormat {
   // Indicates TFRecords format files with gzip compression.
   FORMAT_TFRECORDS_GZIP = 5;
 
-  reserved 1 to 4, 6 to max;
+  // Indicates parquet format files in any of the supported compressions.
+  // https://arrow.apache.org/docs/python/parquet.html#compression-encoding-and-file-compatibility
+  FORMAT_PARQUET = 6;
+
+  reserved 1 to 4, 7 to max;
```
1025KB, line 132: let's use 16
reserved 1 to 4, 7 to 15, 17 to max;

btw, is parquet's file format and payload format both named "Parquet"? I wonder if there are better names to distinguish them?

1025KB (approved):

rcrowe-google (approved):
I think that the changes suggested by @1025KB probably make sense, but otherwise LGTM, and thanks!

martinbomio:
@iindyk @rcrowe-google @1025KB addressed the latest suggestions. Should be ready to merge. I need one of you to do the merging :)

iindyk:
I don't have access to merge here

martinbomio:
@1025KB build seems to be failing with what I think is an unrelated error. Would you mind taking a look?

1025KB:
Hi, Venkat, could you help on merging this PR?

iindyk (approved):

jiyongjung0 (approved):

jiyongjung0:
@martinbomio The build failure seems like related to https://github.com/tensorflow/tfx/commit/517441dbed44467ab20d66da1dd9a3ca3896ba62. Could you rebase the PR and retry?

martinbomio:
@jiyongjung0 done!

jiyongjung0:
@martinbomio It seems like the current failure is a real problem. Could you take a look?

martinbomio:
@jiyongjung0 fixed the issue, can we trigger another build?

tfx/components/util/tfxio_utils.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """TFXIO (standardized TFX inputs) related utilities."""
+from __future__ import annotations
```
martinbomio, line 15: had to add this to allow for proto enums to be used as types in signatures. Not sure if this is something desirable from the tfx side or not @jiyongjung0 @1025KB

tfx/components/util/tfxio_utils_test.py
```diff
@@ -306,7 +307,7 @@ def test_get_tf_dataset_factory_from_artifact(self):
     dataset_factory = tfxio_utils.get_tf_dataset_factory_from_artifact(
         [examples], _TELEMETRY_DESCRIPTORS)
     self.assertIsInstance(dataset_factory, Callable)
-    self.assertEqual(tf.data.Dataset,
+    self.assertEqual('tf.data.Dataset',
```
martinbomio, line 310: this stringify of the signature is a result of adding: `from __future__ import annotations`

tfx/components/util/tfxio_utils.py
martinbomio: I can go back to string in the `make_tfxio ` function signature instead of the proto enum type.

1025KB:
I created an internal pending change and I will make sure it submit (your PR will automatically close once I submit it)

martinbomio:
Hi @rcrowe-google @iindyk @1025KB any update on this PR

1025KB:
the reviewer is just back from vacation, change should be in this week

PR #4803, opened 2022-04-07
Update tft.md

pindinagesh:
tft.md has a broken link, updated with the correct link
(https://www.tensorflow.org/tfx/transform/get_started)

### DIFF
--- a/docs/guide/tft.md
+++ b/docs/guide/tft.md
@@ -2,7 +2,7 @@
 
 Transform is available as a standalone library.
 
--   [Getting Started with TensorFlow Transform](/tfx/transform/get_started)
+-   [Getting Started with TensorFlow Transform](https://www.tensorflow.org/tfx/data_validation/get_started/)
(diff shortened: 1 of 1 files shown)


docs/guide/tft.md
```diff
@@ -2,7 +2,7 @@
 
 Transform is available as a standalone library.
 
--   [Getting Started with TensorFlow Transform](/tfx/transform/get_started)
+-   [Getting Started with TensorFlow Transform](https://www.tensorflow.org/tfx/data_validation/get_started/)
```
jiyongjung0, line 5: Thank you for pointing this out.
Shouldn't it be https://www.tensorflow.org/tfx/transform/get_started instead?

pindinagesh: Sorry, it's my bad that's the right one

jiyongjung0 (approved):
Thanks!!

PR #4816, opened 2022-04-13
Update vertex_pipelines_bq.ipynb

pindinagesh:
Fixing broken link for BigQuery ExampleGen, Updated with correct link
(https://www.tensorflow.org/tfx/api_docs/python/tfx/v1/extensions/google_cloud_big_query/BigQueryExampleGen)


jiyongjung0 (commented):
Thank you for the PR! But the url seems like a domain for china only. Could you use `https://www.tensorflow.org/tfx/api_docs/python/tfx/v1/extensions/google_cloud_big_query/BigQueryExampleGen` instead?

pindinagesh:
Hi@jiyongjung0 
I couldn't find a domain but we can use the url which you mentioned. Thanks

pindinagesh:
Hi @jiyongjung0 

I have updated with it which you mentioned. Please confirm if it looks good. Thanks

jiyongjung0 (approved):
Thanks!

PR #4825, opened 2022-04-24
Update beam_orchestrator.md

pindinagesh:
Fixing broken link in beam_orchestrator.md file with correct link(https://github.com/tensorflow/tfx/blob/master/tfx/examples/chicago_taxi_pipeline/taxi_pipeline_local.py)


jiyongjung0:
Thank you for catching this! However, this document itself is quite out-of-date and need to be updated. We don't recommend using beam as an orchestration engine any more. `LocalDagRunner` is a replacement and it seems like we need to replace this doc with some introduction to `LocalDagRunner`.

Let me track it in a separate thread.

PR #4891, opened 2022-05-23
fix penguin_simple docs: correct for improved readability

jeosol:
After this change, the phrase  "... thought of an deployment process ..." becomes "... thought of as a deployment process ..."


jiyongjung0 (approved):
Thanks!

PR #5093, opened 2022-07-29
Verify output artifacts as found from cache before making execution decision

micahjsmith:
If cache is enabled, and if matching artifacts are found in the cache for a component's outputs, verify the output artifacts before making an execution decision. This avoids a situation in which the artifacts are missing (no object exists at the URI associated with the artifact due to a deletion process outside of the pipeline) and the downstream component that uses this component's outputs fails.

* since the downstream consumer of the cached artifacts will use `BaseDriver.verify_input_artifacts` to verify the artifacts, use the same method when checking the cache to "fail fast" (avoid using the cache)
* note that MLMD ArtifactState is not used at all in the BaseDriver logic. perhaps in a follow-up PR, we can add (1) checks when getting artifacts from cache that the objects are marked LIVE/PUBLISHED and (2) for artifacts from cache that are deemed to be missing given the logic in the current PR, update their state in MLMD to MISSING

cc @1025KB to follow up on our discussion on Wednesday

### DIFF
--- a/tfx/dsl/components/base/base_driver.py
+++ b/tfx/dsl/components/base/base_driver.py
@@ -41,8 +41,8 @@ def _generate_output_uri(base_output_dir: str,
 def _prepare_output_paths(artifact: types.Artifact):
   """Create output directories for output artifact."""
   if fileio.exists(artifact.uri):
-    msg = 'Output artifact uri %s already exists' % artifact.uri
-    absl.logging.warning(msg)
+    absl.logging.warning(
+      'Output artifact uri %s already exists', artifact.uri)
     # TODO(b/158689199): We currently simply return as a short-term workaround
     # to unblock execution retires. A comprehensive solution to guarantee
     # idempotent executions is needed.
@@ -88,9 +88,9 @@ class BaseDriver:
     for single_artifacts_list in artifacts_dict.values():
       for artifact in single_artifacts_list:
         if not artifact.uri:
-          raise RuntimeError('Artifact %s does not have uri' % artifact)
+          raise RuntimeError(f'Artifact {artifact} does not have uri')
         if not fileio.exists(artifact.uri):
-          raise RuntimeError('Artifact uri %s is missing' % artifact.uri)
+          raise RuntimeError(f'Artifact uri {artifact.uri} is missing')
 
   def _log_properties(self, input_dict: Dict[str, List[types.Artifact]],
                       output_dict: Dict[str, List[types.Artifact]],
@@ -145,11 +145,11 @@ class BaseDriver:
             # 0.
             if not artifact.uri or not artifact.id:
               raise ValueError((
-                  'Unresolved input channel %r for input %r was passed in '
+                  'Unresolved input channel {repr(artifact)} for input {repr(name)} was passed in '
                   'interactive mode. When running in interactive mode, upstream '
                   'components must first be run with '
                   '`interactive_context.run(component)` before their outputs can '
-                  'be used in downstream components.') % (artifact, name))
+                  'be used in downstream components.'))
             artifacts_by_id.update({a.id: a for a in artifacts})
         else:
           artifacts = self._metadata_handler.search_artifacts(
@@ -312,7 +312,8 @@ class BaseDriver:
           execution_state=metadata.EXECUTION_STATE_CACHED,
           contexts=contexts)
     else:
-      absl.logging.debug('Cached results not available, move on to new execution')
+      absl.logging.debug(
+        'Cached results not available, move on to new execution')
       # Step 4a. New execution is needed. Prepare output artifacts.
       output_artifacts = self._prepare_output_artifacts(
           input_artifacts=input_artifacts,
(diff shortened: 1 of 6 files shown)


tfx/dsl/components/base/base_driver.py
```diff
@@ -292,7 +292,17 @@ def pre_execution(
...
-    if output_artifacts is not None:
+
+      # Check that cached output artifacts will actually be considered a cache hit by downstream
+      # components
+      if output_artifacts is not None:
+        try:
+          self.verify_input_artifacts(artifacts_dict=output_artifacts)
+          use_cached_results = True
+        except RuntimeError:
+          use_cached_results = False
```
1025KB, line 303: let's log a info message here for debugging reason

1025KB (commented):
Thanks!!

1025KB (commented):
Note that [here](https://github.com/tensorflow/tfx/blob/master/tfx/orchestration/portable/cache_utils.py#L173) needs similar updates too.

base_driver is the old path, new orchestrator like LocalDagRunner uses above portable code

micahjsmith:
> Note that [here](https://github.com/tensorflow/tfx/blob/master/tfx/orchestration/portable/cache_utils.py#L173) needs similar updates too.
> 
> base_driver is the old path, new orchestrator like LocalDagRunner uses above portable code

thanks for the review, will make those changes in a bit!

micahjsmith:
@1025KB I'm looking at the differences between base_driver and cache_utils. I was hoping that base_driver could be refactored to use cache_utils (in the same way it is used by tfx.orchestration.portable.launcher), but it seems there are some significant differences, such as the way they compute cache keys from context and the steps in get_cached_context. unless you tell me otherwise (i.e. that I should look further into refactoring them together), I'm just going to copy/paste my new logic into cache_utils

1025KB:
yep, it's different code path at this point, so logic duplication sounds good. Thanks!!

micahjsmith:
@1025KB ready for re-review!

* refactored the logic of checking whether artifacts exist into `artifact_utils.verify_artifacts`
* use `artifact_utils.verify_artifacts` in three locations: `BaseDriver.verify_input_artifacts`, `BaseDriver.pre_execution`, and `cache_utils.get_cached_outputs`
* add tests
* fix all extant lint issues in touched files, sorry for the large diff as a result!

note that the logic differs between `cache_utils.get_cached_outputs` and `BaseDriver.pre_execution`/`metadata.get_cached_outputs`. in the former case, all the matching executions are checked for valid artifacts. in the latter case, only the most recent matching execution is checked for valid artifacts.

tfx/orchestration/portable/cache_utils.py
```diff
@@ -166,11 +166,15 @@ def get_cached_outputs(
...
   # Defensively traverses candidate executions and returns once we find an
-  # execution with valid outputs.
+  # execution with valid outputs that can be confirmed to still exist.
   for execution in cached_executions:
     cached_output_artifacts = _get_outputs_of_execution(metadata_handler,
                                                         execution.id)
     if cached_output_artifacts is not None:
-      return cached_output_artifacts
+      try:
+        artifact_utils.verify_artifacts(cached_output_artifacts)
```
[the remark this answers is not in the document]
micahjsmith: added!

tfx/orchestration/portable/cache_utils_test.py
```diff
@@ -24,9 +24,10 @@
 from tfx.proto.orchestration import pipeline_pb2
 from tfx.types import standard_artifacts
 from tfx.utils import test_case_utils
+
 from google.protobuf import text_format
 from ml_metadata.proto import metadata_store_pb2
-
+import tensorflow as tf
```
1025KB, line 30: import tf is quite heavy, let's drop it if not necessary?

1025KB, line 30: keep previous order?

rcrowe-google (approved):
LGTM, thanks!

tfx/types/artifact_utils_test.py
```diff
@@ -179,6 +179,21 @@ def testIsArtifactVersionOlderThan(self):
     self.assertFalse(
         artifact_utils.is_artifact_version_older_than(examples, '0.1'))
 
+  @mock.patch("tfx.types.artifact_utils.fileio", autospec=True)
```
jiyongjung0, line 182: How about adding some test cases for the error case? For example uri is missing or the file doesn't exist.

jiyongjung0 (commented):
Thank you for the great contribution!

tfx/orchestration/portable/cache_utils_test.py
micahjsmith: done!

thought I understood the pattern as stdlib, firstpart, thirdparty which would place tf in thirdparty group. but it's not consistent in other files, so not my problem 🤷

tfx/types/artifact_utils_test.py
micahjsmith: added!

1025KB (approved):

rcrowe-google:
It seems that we always assume that artifact payloads are in file systems.  Should we?

1025KB:
> It seems that we always assume that artifact payloads are in file systems. Should we?

at this point it's still the case

PR #5096, opened 2022-08-01
Fix ranking pipeline and add grpcio constraint

rtg0795:
* Skips ranking pipeline tests when struct2tensor is not available.
* Temporarily avoid grpcio==1.48.0 which causes a hang in the penguin_pipeline_local_infraval_e2e_test.py.
*  Add markupsafe constraint

### DIFF
--- a/test_constraints.txt
+++ b/test_constraints.txt
@@ -8,6 +8,9 @@
 # constraint if it is installed either directly or transitively by the
 # dependencies.
 
+# TODO(b/220800338): Workaround for markupsafe.
+markupsafe<2.1.0
+
 # TODO(b/232490018) Remove pinned Cython after pycocotools issue resolved.
 Cython<1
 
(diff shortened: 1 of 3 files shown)


test_constraints.txt
```diff
@@ -8,5 +8,8 @@
 # constraint if it is installed either directly or transitively by the
 # dependencies.
 
-# TODO(b/220800338): Workaround for markupsafe.
```
jiyongjung0, line 11: Could you revive this markupsafe constraint?
It was changed in https://github.com/tensorflow/tfx/commit/b021a486c9363147eca35e97742b555a5ad1cce4 but I don't want to include it in 1.8.x if possible.

rtg0795: Sure jiyong. Can you please review again. Thanks.

jiyongjung0 (approved):
Thanks!

Issue #4434, opened 2021-11-05
Disabling pylint on repo


casassg:
Hi folks!

So I come to realize that Google internally does not run pylint. This is okay, but it's making external contributions painful. I would propose to get rid of the pylint check in this repo as it usually fails from existing code in the repo that was merged w copybara that skips CI entirely. 

That would allow contributors to actually get good signal.

I'm happy to perform the PR myself.

casassg:
CC @rcrowe-google

MarkDaoust:
It can be hard to keep internal and external presubmits in sync.

IIUC Google uses pylint internally, it's just set to only warn about **new** errors added by a change.

casassg:
@MarkDaoust could we then turn the CI job as warning instead? It currently fails and blocks from continuing testing which kinda makes contributors feel forced to fix internal code that is not related to their change which should never be the case

polydier1:
+1 on @MarkDaoust , Google uses the pylint internally -- but we use certain annotations provided by Google pylint for certain cases.

I would vote to make the external lint check as warning to unblock development. Any opinion on this? @1025KB

jiyongjung0:
I'm sorry for the late follow-up. Actually we had some discussion w/ Tensorflow repo maintainers and they also recommended to disable the linter. Let me try changing it to warnings.

Thank you for the suggestions!

singhniraj08:
@casassg,

linter failure is changed to warnings and github CI action will continue to run even if linter fails as per above [comment](https://github.com/tensorflow/tfx/issues/4434#issuecomment-1209218191). 
Kindly let us know if this issue can be closed. Thank you!

jiyongjung0:
I think that this can be closed now. Please reopen if anyone have more suggestion. Thanks!

PR #5124, opened 2022-08-09
Continues github CI action even if linters fail.

jiyongjung0:
Current linter has many false positives and it seems not justify blocking PRs.
We will proceed to next steps and make the entire job as a success if the linter fails.

Reference: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

Please see also https://github.com/tensorflow/tfx/issues/4434


jiyongjung0:
@1025KB Could you take a look at this PR?

rcrowe-google (approved):
If others are comfortable with this then I'm ok too.

PR #5178, opened 2022-08-29
Initial iteration for factory component

casassg:
(please ignore for now, just small iterations on factory component)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,6 +111,7 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+.DS_Store
 
 # Spyder project settings
 .spyderproject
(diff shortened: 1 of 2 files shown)


casassg:
Hey @charlesccychen, @yanyuchka and I started playing around to see if we could make component decorator become a factory. This does not change much the behaviour and still maintains everything working, but it should enable the rest of the work needed for the proposal.

We mostly just moved the executor variables to not be generated at compile time, and instead introspect to generate it at runtime. This is needed in order to later add dynamic spec at component instantiation/pipeline compile time. The executor needs to be agnostic of what the spec is in order to process arguments in different ways depending on how the component gets instantiated.

rcrowe-google:
Still in draft mode, but in my view, so far so good!

PR #5234, opened 2022-09-19
Uses the latest available version when building a container image bas…

rtg0795:
…ed on DLVM images.

Also caps google-api-core version to avoid 1.33.0 which is not compatible with tensorboard.

And deleted a workaround for ill-formed TF 2.9 image because we have a new image without the problem.

PiperOrigin-RevId: 475214110


jiyongjung0 (approved):
Thanks!

PR #5497, opened 2022-11-16
feat: make experimental.annotations.OutputDict compatible with typing.TypedDict

yanyuchka:
With this change, `OutputDict` is now a `typing.TypedDict`; the old syntax is still supported.

Now, we can annotate our `@component` return type as follows:
```python
from typing import TypedDict

@component
def MyValidationComponent(...) -> TypedDict("MyTypedDictOutput", accuracy=float):
   ...
  return {
    'accuracy': accuracy
  }
 ```
Alternative `TypedDict` syntax is supported as well.
The main advantage is that it's TFX-agnostic type that type-checkers can validate.
 
 -----
 ### TODO:
 - [ ] Update any docs/tutorials

### DIFF
--- a/tfx/dsl/component/experimental/annotations.py
+++ b/tfx/dsl/component/experimental/annotations.py
@@ -17,9 +17,17 @@ Experimental. No backwards compatibility guarantees.
 """
 
 import inspect
+import sys
+
 from typing import Any, Type, Union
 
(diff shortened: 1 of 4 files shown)


yanyuchka:
re-opened with different username #5498

Issue #5518, opened 2022-11-22
Running the classify_local.sh crashes the tensorflow serving container


ziadloo:
**System information**

- Have I specified the code to reproduce the issue (Yes, No): **Yes**
- Environment in which the code is executed (e.g., Local(Linux/MacOS/Windows),
  Interactive Notebook, Google Cloud, etc): **Local (Linux)**
- TensorFlow version: **2.9.3**
- TFX Version: **1.10.0**
- Python version: **3.8.15**
- Python dependencies (from `pip freeze` output):

```
absl-py==1.3.0
anyio==3.6.2
apache-beam==2.43.0
argon2-cffi==21.3.0
argon2-cffi-bindings==21.2.0
astunparse==1.6.3
attrs==21.4.0
backcall==0.2.0
beautifulsoup4==4.11.1
bleach==5.0.1
cachetools==4.2.4
certifi==2022.9.24
cffi==1.15.1
charset-normalizer==2.1.1
click==7.1.2
cloudpickle==2.2.0
crcmod==1.7
debugpy==1.6.3
decorator==5.1.1
defusedxml==0.7.1
dill==0.3.1.1
docker==4.4.4
docopt==0.6.2
entrypoints==0.4
fastavro==1.7.0
fasteners==0.18
fastjsonschema==2.16.2
flatbuffers==1.12
gast==0.4.0
google-api-core==1.32.0
google-api-python-client==1.12.11
google-apitools==0.5.31
google-auth==1.35.0
google-auth-httplib2==0.1.0
google-auth-oauthlib==0.4.6
google-cloud-aiplatform==1.19.0
google-cloud-bigquery==2.34.4
google-cloud-bigquery-storage==2.13.2
google-cloud-bigtable==1.7.3
google-cloud-core==2.3.2
google-cloud-datastore==1.15.5
google-cloud-dlp==3.9.2
google-cloud-language==1.3.2
google-cloud-pubsub==2.13.11
google-cloud-pubsublite==1.6.0
google-cloud-recommendations-ai==0.7.1
google-cloud-resource-manager==1.6.3
google-cloud-spanner==3.23.0
google-cloud-storage==2.6.0
google-cloud-videointelligence==1.16.3
google-cloud-vision==1.0.2
google-crc32c==1.5.0
google-pasta==0.2.0
google-resumable-media==2.4.0
googleapis-common-protos==1.57.0
grpc-google-iam-v1==0.12.4
grpcio==1.51.0
[...]

PR #5530, opened 2022-11-24
Replacing missing features in chicago_taxi_clients.py with their statistical attributes

ziadloo:
Using the Tensorflow data validation package, the dataset's statistics are calculated. And then when composing a request payload, in case the record is missing a feature, the value is replaced with one of the following:

If the feature is of type BYTES (categorical), the most common value will be used instead. If the feature is numeric, the average value of the column will be used. The average will also be cast to INT if the feature type is an integer.

This PR is to address a fix for this issue:

https://github.com/tensorflow/tfx/issues/5518

### DIFF
--- a/tfx/examples/chicago_taxi_pipeline/serving/chicago_taxi_client.py
+++ b/tfx/examples/chicago_taxi_pipeline/serving/chicago_taxi_client.py
@@ -31,6 +31,7 @@ from tensorflow_transform.tf_metadata import schema_utils
 from google.protobuf import text_format
 from tensorflow.python.lib.io import file_io  # pylint: disable=g-direct-tensorflow-import
 from tensorflow_metadata.proto.v0 import schema_pb2
+import tensorflow_data_validation as tfdv
 
 _LOCAL_INFERENCE_TIMEOUT_SECONDS = 5.0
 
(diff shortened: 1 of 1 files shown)


tfx/examples/chicago_taxi_pipeline/serving/chicago_taxi_client.py
```diff
@@ -143,7 +147,18 @@ def _do_inference(model_handle, examples_file, num_examples, schema):
...
+        if feature.type == schema_pb2.FLOAT:
+          one_example[name] = [feature_stats.num_stats.mean]
+        elif feature.type == schema_pb2.INT:
+          one_example[name] = [int(feature_stats.num_stats.mean)]
+        elif feature.type == schema_pb2.BYTES:
+          top_values = list(feature_stats.string_stats.top_values)
+          if len(top_values) > 0:
+            one_example[name] = [top_values[0].value.encode('utf8')]
+          else:
+            one_example[name] = ["".encode('utf8')]
```
jiyongjung0, line 161: Could you use single quotes for consistency?

jiyongjung0 (commented):
Thank you so much for the great solution!

Please fix a small issue, then I'll try to merge this.

tfx/examples/chicago_taxi_pipeline/serving/chicago_taxi_client.py
ziadloo: Please let me know if I need to create a new PR or what. This is my first time contributing and I don't know how to proceed. The requested change is made and pushed to my fork and I can see it below. But I don't know if I need to do something before it is part of the PR.

jiyongjung0: This should be enough. Let me begin the internal review process!

jiyongjung0 (approved):

PR #5549, opened 2022-11-30
Display tfma output

jwiklund:
The render_slicing_metrics only returns a view, without either returning or displaying it it won't ever show up.

Also needed https://github.com/tensorflow/model-analysis/pull/165 to test this in jupyter lab.

Edit: I noticed that there is a difference in behavior in colab vs jupyter lab rendering, colab doesn't return anything while jupyter does, so I guess this could break in colab?

### DIFF
--- a/tfx/orchestration/experimental/interactive/standard_visualizations.py
+++ b/tfx/orchestration/experimental/interactive/standard_visualizations.py
@@ -81,7 +81,7 @@ class ModelEvaluationVisualization(visualizations.ArtifactVisualization):
     tfma_result = tfma.load_eval_result(artifact.uri)
     # TODO(ccy): add comment instructing user to use the TFMA library directly
     # in order to render non-default slicing metric views.
-    tfma.view.render_slicing_metrics(tfma_result)
+    display(tfma.view.render_slicing_metrics(tfma_result))
 
 
 class SchemaVisualization(visualizations.ArtifactVisualization):


1025KB:
Could you test it in Colab to ensure? Thanks

rcrowe-google:
>I noticed that there is a difference in behavior in colab vs jupyter lab rendering, colab doesn't return anything while jupyter does, so I guess this could break in colab?

We need to resolve that before we can merge this.  @MarkDaoust @markmcd - Any thoughts on the Colab question?

MarkDaoust:
> Any thoughts on the Colab question?

Just that @jwiklund is correct when he says:

> The render_slicing_metrics only returns a view, without either returning or displaying it it won't ever show up.

> Could you test it in Colab to ensure?

I can't find a single notebook that uses this. But plenty that use `tfma.view.render_slicing_metrics`

rcrowe-google:
It looks to me like that PR was never finished.  It needs to work in both Jupyter and Colab, so I would suggest testing the environment for Colab and switching.

`if 'google.colab' in sys.modules:`

rcrowe-google:
https://github.com/tensorflow/model-analysis/pull/165 was merged, so that should help with this.

briron (approved):

briron:
We couldn't reproduce this issue. It's working well in Colab. Could you provide more context?

rcrowe-google (approved):
Approved following testing by @briron

PR #5710, opened 2023-02-08
Fixed bug where docker does not mount path because it is relative

pritamdodeja:
Docker doesn't mount the model_path because it is relative.  Updated it so that it uses the absolute path.

### DIFF
--- a/tfx/components/infra_validator/model_server_runners/local_docker_runner.py
+++ b/tfx/components/infra_validator/model_server_runners/local_docker_runner.py
@@ -116,7 +116,7 @@ class LocalDockerRunner(base_runner.BaseModelServerRunner):
     if isinstance(self._serving_binary, serving_bins.TensorFlowServing):
       is_local = os.path.isdir(self._model_path)
       run_params = self._serving_binary.MakeDockerRunParams(
-          model_path=self._model_path,
+          model_path=os.path.abspath(self._model_path),
           needs_mount=is_local)
     else:
       raise NotImplementedError('Unsupported serving binary {}'.format(


pritamdodeja:
@gbaned The CLA is signed on the Intuitive side now.  Can you please review this PR and advise once Google countersigns?  Thank you very much! -Pritam

chongkong:
I was looking into the callers and I couldn't find a case the model_path becomes a relative path unless the `artifact.uri` itself is a relative path (which I consider is a malformed artifact uri). Can you provide more context when it become a relative path, so I can tell what is the right fix for this?

pritamdodeja:
> I was looking into the callers and I couldn't find a case the model_path becomes a relative path unless the `artifact.uri` itself is a relative path (which I consider is a malformed artifact uri). Can you provide more context when it become a relative path, so I can tell what is the right fix for this?

My use case is to create the pipeline in a git repo and have it be portable to other machines, hence the relative path usage.

chongkong:
If I understood correctly then your artifact path is a relative path right? I think that's not quite the right use case, and the artifact URI should be unique and persistent.

At least your host and container path can be matched by mounting onto the same name directory, or always use the container path as the source of truth; In either case, artifact URI should not be a relative path because its serialized form can be access anywhere and the content should be consistent.

pritamdodeja:
> If I understood correctly then your artifact path is a relative path right? I think that's not quite the right use case, and the artifact URI should be unique and persistent.
> 
> At least your host and container path can be matched by mounting onto the same name directory, or always use the container path as the source of truth; In either case, artifact URI should not be a relative path because its serialized form can be access anywhere and the content should be consistent.

This is very helpful. Big picture I'm trying to solve how to do development using tfx locally in a distributed fashion, and then push it to vertex for training at scale. Let me know if you require anything else, and whether you want me to close out this PR.  Thank you!

briron:
Would you please close this PR?

PR #5694, opened 2023-02-02
Add SCaNN support to BulkInferrer

IzakMaraisTAL:
In order to load a saved model with a SCaNN layer, we need to import tensorflow_recommenders to avoid a runtime error.

The error would be: Op type not registered 'Scann>ScannSearchBatched'. Note that if you are loading a saved graph which used ops from tf.contrib, accessing (e.g.) `tf.contrib.resampler` should be done before importing the graph, as contrib ops are lazily registered when the module is first accessed.

Since SCaNN is a custom op, after running `pip install scann` the lazy loading mentioned in the error message still needs to be solved.

See https://github.com/google-research/google-research/tree/master/scann

PR #5765, opened 2023-03-01
Add int64 support to bulk-inferrer output types

IzakMaraisTAL:
This is needed to use the bulk-inferrer with a `tensorflow_recommenders.layers.factorized_top_k.TopK`. The TopK layer's `call()` returns a tuple of `(top candidate scores, top candidate identifiers)`. Since the candidate identifiers are integer, the bulk-inferrer must support integer output. Without this support the integer ids are cast to floats, which can cause errors due to loss of precision for large ids.

See https://www.tensorflow.org/recommenders/api_docs/python/tfrs/layers/factorized_top_k/TopK

Related PR: https://github.com/tensorflow/tfx/pull/5694


lego0901 (approved):
Thanks!

PR #5772, opened 2023-03-02
fix: upgrade kfp dependency to >=1.8.14 to fix Google IAP OOB flow

davidxia:
Currently installing `tfx[kfp]` can pull in `kfp` versions older than 1.8.14. This results in authentication errors when used with Google IAP clients that use the
[discontinued Google Oauth out-of-band flow][1].

`kfp` 1.8.14 replaces Google Oauth OOB with the Google [localhost IP address flow][2] which will still work. [See commit history here][3].

> You do not need to do anything related to this deprecation if
> you are using the loopback IP address flow on a Desktop app
> OAuth client as usage with that OAuth client type will
> continue to be supported.

[1]: https://developers.google.com/identity/protocols/oauth2/resources/oob-migration
[2]: https://developers.google.com/identity/protocols/oauth2/resources/loopback-migration
[3]: https://github.com/kubeflow/pipelines/commits/1.8.14

### DIFF
--- a/tfx/dependencies.py
+++ b/tfx/dependencies.py
@@ -132,7 +132,7 @@ def make_extra_packages_airflow():
 def make_extra_packages_kfp():
   """Prepare extra packages needed for Kubeflow Pipelines orchestrator."""
   return [
-      'kfp>=1.8.5,<2',
+      'kfp>=1.8.14,<2',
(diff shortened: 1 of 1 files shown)


daikeshi (approved):

roseayeon (approved):
Thanks for your contribution!

PR #5826, opened 2023-04-01
migrate to mlmd filter query

abinthomasonline:
Migrate to MLMD filter query at two places in the source code.

1. https://github.com/tensorflow/tfx/blob/master/tfx/orchestration/portable/input_resolution/channel_resolver.py#L118
2. https://github.com/tensorflow/tfx/blob/master/tfx/orchestration/portable/mlmd/execution_lib.py#L501

Iteratively querying for each context is causing performance issues for old pipelines. Contexts like "pipeline" and "node" have too many records for old, long-running pipelines in the table.

I found [this](https://github.com/tensorflow/tfx/blob/master/tfx/orchestration/experimental/core/task_gen_utils.py#L245) utility to get executions using filter queries, but this method takes "node" as an argument. Rewrote the method and its invocations to use "contexts" instead.

### DIFF
--- a/tfx/orchestration/experimental/core/async_pipeline_task_gen.py
+++ b/tfx/orchestration/experimental/core/async_pipeline_task_gen.py
@@ -194,7 +194,8 @@ class _Generator:
     # generates a task from it.
     # TODO(b/275231956) Too many executions may have performance issue, it is
     # better to limit the number of executions.
-    executions = task_gen_utils.get_executions(metadata_handler, node)
+    executions = task_gen_utils.get_executions(
+        metadata_handler, node.contexts.contexts)
     oldest_active_execution = task_gen_utils.get_oldest_active_execution(
         executions)
     if oldest_active_execution:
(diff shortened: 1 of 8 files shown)


chongkong:
Thanks Abin for your contribution.

Unfortunately due to the dependency order, we should be using `tfx.orchestration.portable.mlmd.execution_lib.get_executions_associated_with_all_contexts` instead of the one in the `task_gen_utils`. Also we were not actively adopting MLMD filter query in OSS codebase as old kubeflow users with old metadata server will suffer from the unsupported error, so we should be guard from such error.

```
def get_executions_associated_with_all_contexts(
    metadata_handler: metadata.Metadata,
    contexts: Iterable[metadata_store_pb2.Context]
) -> List[metadata_store_pb2.Execution]:
  try:
    # new filter query logic (same as task_gen_utils)
  except mlmd.errors.UnimplementedError:
    logging.warning('MLMD filter query is not supported. Using a slower mechanism.')
    pass

  # Original bad performing code for backward compatibility.
  executions_dict = ...
```

gbaned:
Hi @abinthomasonline Can you please check @chongkong's comments and keep us posted ? Thank you!

PR #5847, opened 2023-04-13
Fixed the broken links in understanding_custom_components.md

tiruk007:
Fixed broken links for the following:
1.learn the core concepts by reading the guide to understanding TFX pipelines 
2.build Python function-based components
3.build container-based components
4.build fully custom components


AnuarTB:
Could youp please confirm that it doesn't work without the change? I was wondering if it should work without the change.

tiruk007:
> Could you please confirm that it doesn't work without the change? I was wondering if it should work without the change.

No, It doesn't work. Please find the below screenshot for reference :
image
 Thank you !

lego0901 (approved):

PR #5858, opened 2023-04-18
Reveal custom_config and custom_executor in BigQuery, CSV, and Query ExampleGens

AlexanderLavelle:
update csv, bigqueryexgen, and queryexgen with custom_config and executor


lego0901:
I believe this PR is a subset of #5860, so would you mind if I disregard this PR and only pay attention into #5860? Thanks!

PR #5689, opened 2023-01-31
Update SchemaGen Executor to natively handle SequenceExample

AlexanderLavelle:
While this PR does not address the need to specify `payload_format` in `ExampleGen` (or `ImportExampleGen`, etc), it does address the disconnect between `ExampleGen`, `SchemaGen`, and `Transform` by specifying tensor representations in `SchemaGen`. This should be a positive benefit to most users as the `tensor representations` will become a natural piece of the pipeline. 

This addresses [5520](https://github.com/tensorflow/tfx/issues/5520), [4714](https://github.com/tensorflow/tfx/issues/4714), and, to a lesser extent, [5361](https://github.com/tensorflow/tfx/issues/5631).

PR #5860, opened 2023-04-18
Allow providing a custom_config with beam ReadFromBigQuery arguments

AlexanderLavelle:


tfx/extensions/google_cloud_big_query/example_gen/executor.py
```diff
@@ -65,9 +68,18 @@ def _BigQueryToExample(pipeline: beam.Pipeline, exec_properties: Dict[str, Any],
   """
   project = utils.parse_gcp_project(exec_properties['_beam_pipeline_args'])
   converter = _BigQueryConverter(split_pattern, project)
+  bq_kwargs = json.loads(exec_properties['custom_config'])
+
+  # if not bq_kwargs:
+  #   bq_kwargs={'dummy':None}
 
   return (pipeline
-          | 'QueryTable' >> utils.ReadFromBigQuery(query=split_pattern)
```
lego0901, line 70: I understand this is a good approach to get rid of redundant dependencies, but we have another code reference of `utils.ReadFromBigQuery` internally. Would you mind to make retain it and add a new argument name `big_query_kwargs`?

tfx/extensions/google_cloud_big_query/example_gen/executor.py
```diff
@@ -65,9 +68,18 @@ def _BigQueryToExample(pipeline: beam.Pipeline, exec_properties: Dict[str, Any],
   """
   project = utils.parse_gcp_project(exec_properties['_beam_pipeline_args'])
   converter = _BigQueryConverter(split_pattern, project)
+  bq_kwargs = json.loads(exec_properties['custom_config'])
+
+  # if not bq_kwargs:
+  #   bq_kwargs={'dummy':None}
```
lego0901, line 74: nit: Remove this

tfx/extensions/google_cloud_big_query/example_gen/executor.py
```diff
@@ -65,9 +68,18 @@ def _BigQueryToExample(pipeline: beam.Pipeline, exec_properties: Dict[str, Any],
   """
   project = utils.parse_gcp_project(exec_properties['_beam_pipeline_args'])
   converter = _BigQueryConverter(split_pattern, project)
+  bq_kwargs = json.loads(exec_properties['custom_config'])
```
lego0901, line 71: Please name the variable more readable, e.g., `big_query_kwargs`

tfx/extensions/google_cloud_big_query/utils.py
```diff
@@ -20,31 +20,8 @@
...
 from apache_beam.options import value_provider
 import tensorflow as tf
-from tfx.utils import telemetry_utils
-
-
-@beam.ptransform_fn
-@beam.typehints.with_input_types(beam.Pipeline)
-@beam.typehints.with_output_types(beam.typehints.Dict[str, Any])
-def ReadFromBigQuery(pipeline: beam.Pipeline,  # pylint: disable=invalid-name
-                     query: str) -> beam.pvalue.PCollection:
```
lego0901, line 33: Maybe changing it into `ReadFromBigQuery(pipeline:beam.Pipeline, query:str, **big_query_custom_config)` would be more appropriate.

tfx/components/example_gen/component.py
```diff
@@ -66,7 +66,8 @@ def __init__(
                                    data_types.RuntimeParameter]] = None,
       output_data_format: Optional[int] = example_gen_pb2.FORMAT_TF_EXAMPLE,
       output_file_format: Optional[int] = example_gen_pb2.FORMAT_TFRECORDS_GZIP,
-  ):
+      custom_executor_spec: Optional[executor_spec.ExecutorSpec] = None,
+    ):
```
lego0901, line 70: nit: Remove 2-spaces indentation from here

tfx/extensions/google_cloud_big_query/example_gen/executor.py
lego0901: Or, if you just retain this function, I will modify the internal references. I agree this is redundant code, too.

tfx/components/example_gen/component.py
lego0901: I think adding `custom_executor_spec` should be split with this PR because it has different purpose of action.
Also, is this change necessary for now? I myself am not confident if there is use cases that `QueryBasedExampleGen` requires `custom_executor_spec` to override the executor spec, as well as `FileBasedExampleGen` provides.

tfx/extensions/google_cloud_big_query/example_gen/executor.py
AlexanderLavelle: @lego0901 I removed the other call of `utils.ReadFromBigQuery` for now

tfx/extensions/google_cloud_big_query/example_gen/executor.py
AlexanderLavelle: gone

tfx/extensions/google_cloud_big_query/example_gen/executor.py
AlexanderLavelle: done

tfx/components/example_gen/component.py
AlexanderLavelle: @lego0901 to the second point -- I anticipated this as well, as I think the inclusion of those bits should be a lower lift. That simplified PR is [here](github.com/tensorflow/tfx/pull/5858) while we discuss the best way to implement the BigQuery pieces

AlexanderLavelle: @lego0901 I'm not sure about the `QueryBasedExampleGen` piece needing the `custom_executor_spec`, but I think the overarching intention here is to provide to average users tools that they can use without more advanced understanding. 

Technically it never makes sense to have `custom_executor_spec` when it appears subclassing accomplishes the same objective?

Another potential fix here would be to add a tutorial on the TFX page to teach users to subclass the components with their custom executor.

AlexanderLavelle: @lego0901 Maybe my first comment here misunderstood what you meant, in which case my apologies :)

tfx/extensions/google_cloud_big_query/example_gen/executor.py
lego0901: I meant, there are internal codebase not exposed to the open source world, and some of them are calling the `utils.ReadFromBigQuery`. Therefore, it will cause some breakages to delete the function definition, so please keep it and I will handle it from my side :)

tfx/components/example_gen/component.py
lego0901: Makes sense. Better to provide than nothing. Ack!

AlexanderLavelle:
@lego0901 I believe I have corrected the issues to a tolerable level. Do you have any more input?

tfx/extensions/google_cloud_big_query/utils.py
```diff
@@ -30,21 +30,21 @@
 @beam.typehints.with_input_types(beam.Pipeline)
 @beam.typehints.with_output_types(beam.typehints.Dict[str, Any])
 def ReadFromBigQuery(pipeline: beam.Pipeline,  # pylint: disable=invalid-name
-                     query: str) -> beam.pvalue.PCollection:
+                     query: str,
+                     big_query_custom_configs: Optional[Dict[str, Any]]) -> beam.pvalue.PCollection:
```
lego0901, line 34: Probably, `big_query_custom_configs: Optional[Dict[str, Any]] = None` would be recommended, but I will do it by myself on the internal review!

lego0901:
LGTM, I will use internal formatter and some minor updates for the internal review.
Thanks for your contribution and hard work!

lego0901 (approved):

AlexanderLavelle:
@lego0901 I realized also that BQEG even with a custom executor for some reason does not handle `tf.train.SequenceExample`s -- so there should also be a passthrough for `payload_format` or `output_data_format`, the latter of which is an arg in `QueryBasedExampleGen` 

I tested this downstream working with (another PR of mine)[https://github.com/tensorflow/tfx/pull/5689] so I can potentially raise an issue since there is a sneaky issue somewhere.



Full explanation of this bug:
Somewhere in the process of encoding the BQEG examples with a custom executor that allows sequence examples there is an issue. The records write as SequenceExamples without problem, I have recovered them and investigated. My SchemaGen executor has been modified to implement the tensor representations, so that isn't the issue. The Transform component has a confusing path of determining the Sequence/Regular Example. I have to make the educated guess that SchemaGen is confused because the payload format artifact can only be / defaults HARD to tf.train.Example / example_pb2.{whatever it is called}

PR #5900, opened 2023-05-10
[Temporary] Add instructions to use previous version of Python

AnuarTB:
As Colab has upgraded its Python version to 3.10, we are expiriencing some issues with some of the libraris which haven't completely migrated. This workaround will help users to run their tutorials while the migration is ongoing.


AnuarTB:
Not needed, closing this PR.

PR #5906, opened 2023-05-11
Refactor Utils to support added type casting capabilities

AlexanderLavelle:
To facilitate @1025KB / `jyzhao`'s `TODO`, separate functionality to support extended capabilities.

Also enables more helpful error message to be clear on which key(s) are causing problems. 
F-strings work in both supported Python versions on PyPi (3.8, 3.9)

Recommended expanded types handling: 
Decimal - to int and float lists (BigQuery sometimes encodes numbers as Decimal)
Dictionaries - get to lists or list-of-lists
Datetimes - to UNIX / epoch time probably

Further discussion but possible and plausible types:
List-of-Lists - for Sequence Examples*

*This utils function could/would/should produce `SequenceExamples`. In my estimation per [PR 5689](https://github.com/tensorflow/tfx/pull/5689), I posit all TFX `ExampleGens` should produce `SequenceExamples` as this is forward-looking (in terms of supporting NLP applications etc) and because a `SequenceExample` with empty `sequence_features` seems not only permissible but also is not world breaking with `SchemaGen`, unlike the situation currently. 

Passing a `SequenceExample` to `SchemaGen` without PR 5689 means there has to be some other piece of code used to make `TF Transform` work -- what average user should be burdened with figuring that out?

### DIFF
--- a/tfx/components/example_gen/utils.py
+++ b/tfx/components/example_gen/utils.py
@@ -128,7 +128,8 @@ def dict_to_example(instance: Dict[str, Any]) -> example_pb2.Example:
       feature[key] = pyval_to_feature(pyval, key)
 
     else:
-      raise RuntimeError(f'Key {key} with Column type {type(value)} is not supported.')
+      raise RuntimeError("""Key {} with Column type {} 
+                         is not supported.""".format(key, type(value)))
     
   return example_pb2.Example(features=feature_pb2.Features(feature=feature))
 
--- a/tfx/components/example_gen/utils.py
+++ b/tfx/components/example_gen/utils.py
@@ -84,26 +84,7 @@ _DEFAULT_ENCODING = 'utf-8'
 _SINGLE_VALUE_TYPES = (int, str, float)
 
 
-def value_to_feature(pyval, key) -> feature_pb2.Feature:
-  if pyval is None:
-    return feature_pb2.Feature()
-  elif isinstance(pyval, int):
-    return feature_pb2.Feature(
-        int64_list=feature_pb2.Int64List(value=[pyval]))
-  elif isinstance(pyval, float):
-    return feature_pb2.Feature(
-        float_list=feature_pb2.FloatList(value=[pyval]))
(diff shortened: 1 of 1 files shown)


AnuarTB:
Started looking at a PR, but we have some error with Basel. Currently, it is known issue and we are investigating it.

tfx/components/example_gen/utils.py
```diff
@@ -101,36 +140,15 @@ def dict_to_example(instance: Dict[str, Any]) -> example_pb2.Example:
...
-            bytes_list=feature_pb2.BytesList(
-                value=[v.encode(_DEFAULT_ENCODING) for v in pyval]))
-      else:
-        raise RuntimeError('Column type `list of {}` is not supported.'.format(
-            type(value[0])))
+      feature[key] = list_to_feature(pyval)
+
     else:
-      raise RuntimeError('Column type {} is not supported.'.format(type(value)))
+      raise RuntimeError(f'Key {key} with Column type {type(value)} is not supported.')
```
AnuarTB, line 150: Can you make the error message in consistent format as the other two.

AnuarTB (approved):
Thanks for the PR. Made a comment.

gbaned (approved):

tfx/components/example_gen/utils.py
AlexanderLavelle: done

AlexanderLavelle:
@gbaned Updated with a simplified version by casting single types to a list to reduce two functions to one

tfx/components/example_gen/utils.py
```diff
@@ -101,36 +121,16 @@ def dict_to_example(instance: Dict[str, Any]) -> example_pb2.Example:
     if isinstance(pyval, bytes):
       pyval = pyval.decode(_DEFAULT_ENCODING)
 
-    if pyval is None:
```
lego0901, line 104: Actually, this case was not handled. So I revived `if pyval is None` conditional. Thanks!

PR #5498, opened 2022-11-16
feat: make experimental.annotations.OutputDict compatible with typing.TypedDict

yanyuchka:
With this change, `OutputDict` is now a `typing.TypedDict`; the old syntax is still supported.

Now, we can annotate our `@component` return type as follows:
```python
from typing import TypedDict

@component
def MyValidationComponent(...) -> TypedDict("MyTypedDictOutput", accuracy=float):
   ...
  return {
    'accuracy': accuracy
  }
 ```
Alternative `TypedDict` syntax is supported as well.
The main advantage is that it's a TFX-agnostic type that type-checkers can validate.
 
 -----
 ### TODO:
 - [ ] Update any docs/tutorials

PR #5982, opened 2023-06-23
feat: make experimental.annotations.OutputDict compatible with typing.TypedDict [Updated]

zyang7:
Same effort as #5498, but updated to adjust for the [deprecation](https://docs.python.org/3/library/typing.html#:~:text=Deprecated%20since%20version%203.11%2C%20will%20be%20removed%20in%20version%203.13%3A%20The%20keyword%2Dargument%20syntax%20is%20deprecated%20in%203.11%20and%20will%20be%20removed%20in%203.13.%20It%20may%20also%20be%20unsupported%20by%20static%20type%20checkers.) of kwargs in TypedDict.

From Olga's PR:

> With this change, OutputDict is now a typing.TypedDict; the old syntax is still supported.

> Now, we can annotate our @component return type as follows:

```
from typing import TypedDict

@component
def MyValidationComponent(...) -> TypedDict("MyTypedDictOutput", accuracy=float):
   ...
  return {
    'accuracy': accuracy
  }
```
> Alternative TypedDict syntax is supported as well.
> The main advantage is that it's a TFX-agnostic type that type-checkers can validate.


zyang7:
I'm not sure why the build step continues to fail, any changes I need to make to get the CI to go through? @lego0901 @gbaned

lego0901:
Let me take a look at the CI failure. Thanks!

lego0901:
Hi @zyang7, I submitted #5991 that make it passes the Github CI test. Could you please rebase the master and re-trigger the  CI? Thank you.

zyang7:
Thank you @lego0901, I rebased but I don't think I have permission to re-run the pipeline (it has been running on a daily schedule).

lego0901 (approved):
This looks good to me, except for some lint errors that can be handled by ourselves.
Let me approve it and test it internally. Thank you!

lego0901:
Thanks @zyang7 for your contribution!

We implemented this PR by ourselves (#6212) because (1) your implementation was breaking our internal-Google tests, (2) and there was no simple fix.

PR #6003, opened 2023-07-07
Default to Sequence Examples

AlexanderLavelle:
This is a draft / suggestion based on [5689](https://github.com/tensorflow/tfx/pull/5689)

Everything now goes to Sequence Examples. There is an isinstance filter to navigate the types. 

This supports long-term development by natively implementing sequence type features, which TFX would need to lean on in order to support LLM training and so on. 

@lego0901

### DIFF
--- a/tfx/components/example_gen/component.py
+++ b/tfx/components/example_gen/component.py
@@ -159,7 +159,7 @@ class FileBasedExampleGen(base_beam_component.BaseBeamComponent):
                                     data_types.RuntimeParameter]] = None,
       range_config: Optional[Union[range_config_pb2.RangeConfig,
                                    data_types.RuntimeParameter]] = None,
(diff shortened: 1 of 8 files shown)


gbaned:
Hi @lego0901 Can you please review this PR ? Thank you!

lego0901:
Hi @AlexanderLavelle, thanks a lot for your efforts on this.

We had a look at your PR and discussed internally. Unfortunately, we decided to politely reject your PR.

The main reason was "the backward compatibility". As a proto message type, `SequenceExample` is not compatible with the `Example`. What it means is that every code that uses `tf.train.Example`, regardless of if it is internal or external-Google, should change their way of packing the messages. We assumed that will bring a big burden to our existing customers to upgrade the TFX version.

I always appreciate your concern to our product. Sorry for my response. Thank you.

PR #6060, opened 2023-07-19
Update pusher executor to use constants for custom property keys

jeongukjae:
Found that `_PUSHED_KEY` is defined, it looks like it is for this, but not used anywhere.
So updating the executor to use that constant for property keys.


AnuarTB:
LGTM. Thank you for contribution!

AnuarTB (approved):

gbaned (approved):

PR #6297, opened 2023-09-18
Remove 'awaiting' labels when user issue/PR updated.

shmishra99:
Remove the label "stat:awaiting response", when issue/PR unstale.


lego0901 (approved):
Thanks!

PR #6718, opened 2024-03-21
Update setup_beam_on_spark.sh

levpachmanov:
Actually fix CVE-2018-17190 - the current code just sets `spark.master.rest.enabled true` leaving `spark.authenticate` to it's default value


roseayeon (approved):
Thanks for your contribution!

PR #6889, opened 2024-08-02
Revamp testing workflow and use `pytest` instead of `unittest`

smokestacklightnin:
Here is a summary of what is included in this PR:

- As part of this workflow, pytest has been added so that we can drop the previous custom test collection system, resolving a long-standing TODO. pytest automatically collects all unittest tests.
- Moving to pytest also allows tests to take advantage of [marks](https://docs.pytest.org/en/stable/example/markers.html), which will make it simple to run different types of tests
- End-to-end tests have been split out into a separate job from unit tests, which should reduce total test time and decrease flakiness
- Fixed a bad import statement for the penguin example in kubeflow_gcp_perf_test.py. @roseayeon @gbaned  can you verify that this is the correct import?
- End-to-end tests look like they require a few extra environment variables which point to a container image in order to actually run. I wasn't able to test these locally. @roseayeon @gbaned is there a container that you've been using to run these e2e tests? When the containers are not present, the tests requiring them are now skipped.
- Run all tests by default, leaving users the option to skip e2e tests with the `-m "not e2e"` flag
- Remove `if __name__ == "__main__"` section from tests because it is not necessary, or even run, when using pytest. Nontrivial functionality from those sections was preserved.
- Remove deprecated `pytest_runner` from dependencies
- Consolidate setup python and cache actions into just setup python action. Upgraded to v5.
- Use [setup-bazel](https://github.com/bazel-contrib/setup-bazel) workflow instead of using curl to download bazel
- Linting was removed from test workflow in anticipation of adding linting in a subsequent PR
- Update `CONTRIBUTING.md` to reflect all changes
- Use a [matrix](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow) to test multiple python versions and both e2e and non-e2e tests, minimizing repeated workflow code.
- Remove `enable_v2_behavior` from tests that previously called it because tensorflow v2 is what is currently used.
- Add `xfail` marks to classes and methods that have failing tests


.github/workflows/ci-test.yml
```diff
@@ -1,39 +1,31 @@
...
-      with:
-        fileOutput: ' '
-    - name: Select files to check
-      run: |
-        # Filter out non-python files.
-        (cat $HOME/files_added.txt; echo; cat $HOME/files_modified.txt) | tr ' ' '\n' | grep '\.py$' > py_files.txt || true
-        # Filter out non-test python files and e2e or integration tests.
-        cat py_files.txt | grep '_test\.py$' | grep -v _e2e_ | grep -v integration | grep -v 'examples/' > py_test_files.txt || true
-        # Select proto files.
-        (cat $HOME/files_added.txt; echo; cat $HOME/files_modified.txt) | tr ' ' '\n' | grep '\.proto$' > proto_files.txt || true
```
peytondmurray, line 31: Thanks for this!

tfx/dependencies.py
```diff
@@ -154,7 +154,8 @@ def make_extra_packages_test():
   # Note: It is okay to pin packages to exact versions in this list to minimize
   # conflicts.
   return make_extra_packages_airflow() + make_extra_packages_kfp() + [
-      'pytest>=5,<7',
+      'pytest>=5,<=8',
+      'pytest-subtests==0.13.1',
```
peytondmurray, line 158: What do we need `pytest-subtests` for?

tfx/examples/penguin/_penguin_kubeflow_gcp.py
peytondmurray, line 1: Can we remove this empty file, or is it needed for some reason?

tfx/orchestration/kubeflow/e2e_tests/kubeflow_gcp_perf_test.py
```diff
@@ -241,7 +246,7 @@ def testFullTaxiGcpPipeline(self):
         'parameterServerCount': self._PARAMETER_SERVER_COUNT
     }
 
-    pipeline = penguin_kubeflow_gcp.create_pipeline(
+    pipeline = _penguin_kubeflow_gcp.create_pipeline(
```
peytondmurray, line 249: It seems like this should fail because `tfx.examples.penguin._penguin_kubeflow_gcp` is empty. Should this instead be referring to `tfx.examples.penguin.penguin_pipeline_kubeflow.create_pipeline`?

tfx/orchestration/kubeflow/test_utils.py
```diff
@@ -387,24 +389,27 @@ class BaseKubeflowTest(test_case_utils.TfxTest):
...
+    _BUCKET_NAME = os.environ['KFP_E2E_BUCKET_NAME']
 
-  # The location of test data. The input files are copied to a test-local
-  # location for each invocation, and cleaned up at the end of test.
-  _TEST_DATA_ROOT = os.environ['KFP_E2E_TEST_DATA_ROOT']
+    # The location of test data. The input files are copied to a test-local
+    # location for each invocation, and cleaned up at the end of test.
+    _TEST_DATA_ROOT = os.environ['KFP_E2E_TEST_DATA_ROOT']
+  except KeyError as err:
+    pytest.skip(f"Environment variable {err} not found.", allow_module_level=True)
```
peytondmurray, line 412: :+1:

peytondmurray (changes requested):
Looks nice, thanks for this! Aside from comments, one question: everywhere there is an `e2e` pytest mark there is a `slow` mark. Can we just get rid of the `slow` mark since it is redundant here?

tfx/dependencies.py
smokestacklightnin: I believe there are `unittest` subtests, and they don't work with `pytest` unless this included

tfx/orchestration/kubeflow/e2e_tests/kubeflow_gcp_perf_test.py
smokestacklightnin: Yes. Fixed.

tfx/examples/penguin/_penguin_kubeflow_gcp.py
smokestacklightnin: Done, c.f. [the incorrect import](https://github.com/tensorflow/tfx/pull/6889#discussion_r1705025141)

peytondmurray (approved):
Looks good. Thanks for this :rocket:

tfx/scripts/tfx_test_installed.sh
```diff
@@ -113,7 +114,8 @@ SKIP_LIST+=(
...
 # failure to exit code 255.
 
+# pytest is used now
+
 set -f  # Disable bash asterisk expansion.
-find src -name '*_test.py' \
-  ${SKIP_LIST[@]/#tfx/-not -path src} \
-  |  xargs -I {} sh -c "${PYTHON_BINARY} {} || exit 255"
+${PYTEST_BINARY} $(find src -name '*_test.py' \
+  ${SKIP_LIST[@]/#tfx/-not -path src} | tr '\n' ' ')
```
peytondmurray, line 121: I think we can get rid of this manual test collection - this is now pytest's job.

tfx/scripts/tfx_test_installed.sh
peytondmurray, line 1: Arguably we should remove this; invoking tests from inside bash scripts is not really standard practice. Furthermore this script installs specific versions of dependencies outside of what is in `setup.py`, which is a recipe for a broken test workflow. In particular installing `tfx` from PyPI as is done below will not test the version in the repo, potentially leading developers to think they are testing the development version and in fact testing stuff installed to `site-packages`.

peytondmurray (changes requested):
Thanks for updating the bazel version. I think we should get rid of `tfx_test_installed.sh`. What do you think?

smokestacklightnin:
> I think we should get rid of tfx_test_installed.sh. What do you think?

I agree in concept. My only hesitation is that it might break something internal for Google. That said, if we are disconnecting from Google's infrastructure (which we are), then this won't break anything and it is not needed

roseayeon (approved):

smokestacklightnin:
Now that I've rebased to include linting, I need to fix linting errors.

lego0901:
Thanks for doing the work!

> Some of the tests fail, but they also failed before this PR. The failures were hidden before this PR because only tests for changed files were run. @roseayeon @gbaned how would you like to handle this?

Could you add some feature to skip the failing tests in this moment to fix it later if everything is settled? And please tag me so that I can keep track of it. Thanks.

.github/workflows/ci-test.yml
```diff
@@ -1,56 +1,51 @@
...
+  USE_BAZEL_VERSION: "6.5.0"
+  # Changed to match tensorflow
+  # https://github.com/tensorflow/tensorflow/blob/master/.bazelversion
 
 jobs:
-  build:
+  tests:
     if: github.actor != 'copybara-service[bot]'
     runs-on: ubuntu-latest
     timeout-minutes: 60
```
smokestacklightnin, line 21: Remove this line

smokestacklightnin:
@lego0901 @roseayeon All checks currently pass and the PR is ready for review and merge. All failing tests have been given [`xfail`](https://docs.pytest.org/en/stable/how-to/skipping.html) marks

lego0901:
Hi @smokestacklightnin sorry for the late reply. Could you point me out where the list of the skipped tests is so that we can handle it later? Thanks!

smokestacklightnin:
> Hi @smokestacklightnin sorry for the late reply. Could you point me out where the list of the skipped tests is so that we can handle it later? Thanks!

Hi @lego0901, no problem at all. To see which tests have `xfail`, run the following command in the repo:

```bash
$ grep "pytest.mark.xfail" -r tfx/
```

smokestacklightnin:
@lego0901 FYI, per the request in the Google/Quansight meeting today (Wednesday), I have added tests for both nightly and default constraints. As I am writing this comment, I am waiting for the tests to pass, and once they do, I believe this PR is ready for merge, provided it passes your review.

roseayeon (approved):

PR #6892, opened 2024-08-06
Add a workflow to build wheels and (only on release) publish to PyPI

peytondmurray:
This PR adds a workflow which automatically builds an sdist and a wheel when a pull request is made. If a release is made, the wheels are built and then automatically published to PyPI. Unfortunately I cannot use `cibuildwheel` to target many different architectures at this time because the build system is quite unusual, and will need some additional work to make it compatible with current best practices.

@pradeepkuppla We'll need to enable trusted publishing on the [`tfx` project page](https://pypi.org/project/tfx/) to get this to work. [Instructions are here](https://github.com/pypa/gh-action-pypi-publish?tab=readme-ov-file#trusted-publishing), or I can guide you through this - it's straightforward to do, and should take only a few minutes.

In testing the build locally, I've also added a few build artifacts to `.gitignore`. [I tested this workflow to the extent that I could](https://github.com/peytondmurray/tfx/actions/runs/10275586921/job/28434484804?pr=1), but I will need to get trusted publishing set up on PyPI for this project to be sure about it.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -1,0 +1,100 @@
+name: Build Wheels & Publish to PyPI
+
+on:
+  pull_request:
+  workflow_dispatch:
+  release:
+    types: [published]
+
+jobs:
+  build_sdist:
+    name: Build sdist
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out the repo
+      uses: actions/checkout@v4
+
+    - name: Set up python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.10'
+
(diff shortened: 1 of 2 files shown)


smokestacklightnin (approved):

smokestacklightnin (approved):

roseayeon (approved):

peytondmurray:
Looks like there are some breaking changes between `actions/upload-artifact` v3 and v4 which [disallow modifying artifacts](https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes).

PR #6894, opened 2024-08-07
Add pre-commit lint action

aktech:
This PR adds a pre-commit lint action. Since there are lots of linting issues when running `ruff`, this actions only runs linting on the changed files in the PR, until we fix all the linting issues.

In future we can also directly use: https://pre-commit.ci/ which would also auto-commit linting issues. I didn't went with it right away as I couldn't find a way to do a selective run on the changed files with that.

The proposal is to get this in and start linting pull requests and slowly fix linting errors and then start running pre-commit on full code based instead of PR diff.

### DIFF
--- a/.github/workflows/ci-lint.yml
+++ b/.github/workflows/ci-lint.yml
@@ -1,0 +1,33 @@
+name: pre-commit
+
+on:
+  pull_request:
+  push:
+     branches: [master]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4.1.7
+      with:
+        # Ensure the full history is fetched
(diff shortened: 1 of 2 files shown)


peytondmurray (approved):
This looks good and I agree, if we turn on e.g. `ruff-format` right now the entire repo will be hit. Probably best to do in a separate PR, or bit-by-bit.

aktech:
This PR does turns ruff on, but only on the PR diffs, as in for the commits in the PR (new code basically), without worrying about anything before that.

roseayeon:
Thanks for your work on this. However, merging this PR directly into our internal codebase could potentially cause issues with our internal systems. To avoid any disruptions, I'll separate the internal code from the GitHub code before approving the PR. Thank you for your understanding.

PR #6907, opened 2024-08-21
Add basic `pyproject.toml`

aktech:
The `setup.py` and package dependencies are quite complicated. It would be non-trivial to replace all with `pyproject.toml` in one PR. This is one of many PRs to surgically remove setup.py and use `pyproject.toml`.

Since `pyproject.toml` is a static configuration file in contrast to dynamic setup.py, we're trying to use both at the moment until we fully replace everything in setup.py.

We're currently building three packages in this project:
- tfx
- ml-pipelines-sdk
- tfx-dev (for development, which is a union of the above two)

I have created explicit `pyproject.toml` files for the three types of builds we're doing. There might be some duplication in these files, but now is not the time to extract them, until we fully replace `setup.py` and old build machinery (including installation of specific components via optional dependencies instead of environment variables).


package_build/ml-pipelines-sdk/pyproject.toml
```diff
@@ -0,0 +1,37 @@
+[build-system]
+requires = ["setuptools>=72", "wheel", "pytest-runner", "tomli"]
```
peytondmurray, line 2: Hmm, what do we need `pytest-runner` for as a build-time dependency?

package_build/tfx/pyproject.toml
```diff
@@ -0,0 +1,37 @@
+[build-system]
+requires = ["setuptools>=72", "wheel", "pytest-runner", "tomli"]
```
peytondmurray, line 2: See other note about `pytest-runner` as a build-time dep.

.github/workflows/ci-test.yml
```diff
@@ -57,7 +57,7 @@ jobs:
         python -m pip install --upgrade pip wheel
         # TODO(b/232490018): Cython need to be installed separately to build pycocotools.
         python -m pip install Cython -c ./test_constraints.txt
-        TFX_DEPENDENCY_SELECTOR=NIGHTLY pip install -c ./test_constraints.txt --extra-index-url https://pypi-nightly.tensorflow.org/simple --pre --editable .[all]
+        TFX_DEPENDENCY_SELECTOR=NIGHTLY pip install -c ./test_constraints.txt --extra-index-url https://pypi-nightly.tensorflow.org/simple --pre --editable .[all] --no-binary :all:
```
peytondmurray, line 60: What does this change do?

peytondmurray (changes requested):
Need a double check on the `pytest-runner` build time dependency.

package_build/ml-pipelines-sdk/pyproject.toml
aktech: >Hmm, what do we need pytest-runner for as a build-time dependency?

Good catch. No it's not required. I missed removing it from project directories.

.github/workflows/ci-test.yml
aktech: The idea was to avoid building wheel for `tfx-dev` as it's [not supported/recommended](https://github.com/tensorflow/tfx/blob/55a076c502c7406c96406acfa5122626f2bb81e2/setup.py#L78) for it 

but that doesn't seems to be working. I'll explore alternatives.

setup.py
```diff
@@ -314,49 +311,21 @@ def run(self):
...
-        'Topic :: Scientific/Engineering :: Artificial Intelligence',
-        'Topic :: Scientific/Engineering :: Mathematics',
-        'Topic :: Software Development',
-        'Topic :: Software Development :: Libraries',
-        'Topic :: Software Development :: Libraries :: Python Modules',
-    ],
     namespace_packages=[],
     install_requires=install_requires,
     extras_require=extras_require,
     # TODO(b/158761800): Move to [build-system] requires in pyproject.toml.
```
peytondmurray, line 322: I think this can now be removed.

pyproject.toml
```diff
@@ -0,0 +1,46 @@
...
+    "Topic :: Software Development :: Libraries :: Python Modules"
+]
+keywords = ["tensorflow", "tfx"]
+requires-python = ">=3.9,<3.11"
+[project.urls]
+Homepage = "https://www.tensorflow.org/tfx"
+Repository = "https://github.com/tensorflow/tfx"
+
+[tool.pytest.ini_options]
+addopts = "--verbose -m 'not end_to_end'"
```
peytondmurray, line 40: I'd vote to remove this line because
- Users should be in control of verbosity
- IMO all tests should run by default

Thoughts?

package_build/initialize.sh
```diff
@@ -28,6 +28,7 @@ do
   ln -sf $BASEDIR/dist $BASEDIR/package_build/$CONFIG_NAME/
   ln -sf $BASEDIR/tfx $BASEDIR/package_build/$CONFIG_NAME/
   ln -sf $BASEDIR/README*.md $BASEDIR/package_build/$CONFIG_NAME/
+  ln -sf $BASEDIR/LICENSE $BASEDIR/package_build/$CONFIG_NAME/
```
peytondmurray, line 31: Do we intend to remove this file in a future revision of the build system? It's unusual to be linking in a bunch of files and directories as part of a build step, and I'm wondering if there's a better way.

OTOH if this is just temporary until the rest of the build system rework can be done I'm on board.

.github/workflows/ci-test.yml
```diff
@@ -57,7 +57,7 @@ jobs:
         python -m pip install --upgrade pip wheel
         # TODO(b/232490018): Cython need to be installed separately to build pycocotools.
         python -m pip install Cython -c ./test_constraints.txt
-        TFX_DEPENDENCY_SELECTOR=NIGHTLY pip install -c ./test_constraints.txt --extra-index-url https://pypi-nightly.tensorflow.org/simple --pre --editable .[all]
+        TFX_DEPENDENCY_SELECTOR=NIGHTLY pip install -c ./test_constraints.txt --extra-index-url https://pypi-nightly.tensorflow.org/simple --pre .[all]
```
peytondmurray, line 60: I'm curious - why is editable no longer necessary? Anyway, this is a major improvement - thank you!

peytondmurray (approved):

AnuarTB (approved):

.github/workflows/ci-test.yml
aktech: It has to do with the building of wheels. When using `--editable`, no wheel is built. Instead, a symbolic link is created, pointing to your source code directory, allowing you to edit the code and see the changes instantly. The building of wheels is not recommended for `tfx-dev` package: https://github.com/tensorflow/tfx/blob/4012ecbe64d737ada5ae68ac8f8ce874bc23aa7a/setup.py#L76-L83

See https://github.com/tensorflow/tfx/blob/4012ecbe64d737ada5ae68ac8f8ce874bc23aa7a/package_build/README.md?plain=1#L49

Now we just skip building wheels (even without `--editable` flag), instead of raising error and stopping the installation altogether.

package_build/initialize.sh
aktech: >Do we intend to remove this file in a future revision of the build system? It's unusual to be linking in a bunch of files and directories as part of a build step, and I'm wondering if there's a better way.

I agree with you and I would prefer to get rid of it as well, but the problem is we're building two packages (one subset of other) with the same source code and there is no way to add multiple `pyproject.toml`(s) in the root directory (or I am not aware of any), hence this linking. So, until we find an alternative to achieve this, unfortunately we have to keep this.

pyproject.toml
aktech: Completely agree, but I would propose to handle it in a separate pull request to isolate and minimize functional changes, this is basically a copy of already existing `setup.cfg` (removed in this PR): https://github.com/tensorflow/tfx/blob/4012ecbe64d737ada5ae68ac8f8ce874bc23aa7a/setup.cfg#L5

PR #6919, opened 2024-09-29
Move docs dependencies to common location and check for broken docs on pull request

smokestacklightnin:
This PR accomplishes two main things

- Move the docs dependencies to `docs-requirements.txt`. Before, the requirements were repeated in `tfx/dependencies.py` and in the github docs workflow.
- Build but do not deploy the docs on pull request. This is to check if any PR will break the docs.

### DIFF
--- a/.github/workflows/cd-docs.yml
+++ b/.github/workflows/cd-docs.yml
@@ -26,6 +26,7 @@ jobs:
           cache-dependency-path: |
             setup.py
             tfx/dependencies.py
+            requirements-docs.txt
 
       - name: Save time for cache for mkdocs
         run: echo "cache_id=$(date --utc '+%V')" >> $GITHUB_ENV
--- a/.github/workflows/cd-docs.yml
+++ b/.github/workflows/cd-docs.yml
@@ -2,8 +2,7 @@ name: deploy-docs
 on:
   workflow_dispatch:
   push:
-    branches:
-      - master
+  pull_request:
 permissions:
   contents: write
 jobs:
@@ -17,6 +16,7 @@ jobs:
         run: |
           git config user.name github-actions[bot]
           git config user.email 41898282+github-actions[bot]@users.noreply.github.com
+        if: (github.event_name != 'pull_request')
 
       - name: Set up Python 3.9
         uses: actions/setup-python@v5
@@ -43,3 +43,8 @@ jobs:
 
       - name: Deploy to GitHub Pages
         run: mkdocs gh-deploy --force
+        if: (github.event_name != 'pull_request')
+
+      - name: Build docs to check for errors
+        run: mkdocs build --verbose
+        if: (github.event_name == 'pull_request')
--- a/.github/workflows/cd-docs.yml
+++ b/.github/workflows/cd-docs.yml
@@ -47,5 +47,5 @@ jobs:
         if: (github.event_name != 'pull_request')
 
       - name: Build docs to check for errors
-        run: mkdocs build --verbose
+        run: mkdocs build
         if: (github.event_name == 'pull_request')
--- a/.github/workflows/cd-docs.yml
+++ b/.github/workflows/cd-docs.yml
@@ -40,7 +40,7 @@ jobs:
             mkdocs-material-
 
       - name: Install Dependencies
-        run: pip install $(<requirements-docs.txt)
+        run: pip install -r requirements-docs.txt
 
       - name: Deploy to GitHub Pages
         run: mkdocs gh-deploy --force
--- a/.github/workflows/cd-docs.yml
+++ b/.github/workflows/cd-docs.yml
@@ -39,7 +39,7 @@ jobs:
             mkdocs-material-
 
       - name: Install Dependencies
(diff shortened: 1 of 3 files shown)


.github/workflows/cd-docs.yml
```diff
@@ -39,7 +40,12 @@ jobs:
             mkdocs-material-
 
       - name: Install Dependencies
-        run: pip install mkdocs mkdocs-material mkdocstrings[python] griffe-inherited-docstrings mkdocs-autorefs mkdocs-jupyter mkdocs-caption markdown-grid-tables
+        run: pip install $(<requirements-docs.txt)
```
peytondmurray, line 43:
```suggestion
        run: pip install -r requirements-docs.txt
```

tfx/dependencies.py
```diff
@@ -262,17 +262,12 @@ def make_extra_packages_docs() -> list[str]:
...
-        "mkdocs-caption",
-        "pymdown-extensions",
-        "markdown-grid-tables",
-    ]
+    with open(Path(__file__).resolve().parent.parent / "requirements-docs.txt", "r") as fp:
+        reqs = fp.readlines()
+
+    reqs = [req.replace("\n", "") for req in reqs]
+
+    return reqs
```
peytondmurray, line 270:
```suggestion
    return [req.replace("\n", "") for req in reqs]
```

Can just return this, I think. Also, doesn't `\n` need to be replaced with `' '`?

In the future, we are now using `pyproject.toml` so if we can it would be useful to convert all dependencies to dynamic metadata to make `setuptools` responsible for loading and handling this (and other dependency groups): https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata

Apparently this will not work for just one optional dependency - all must be specified dynamically - but anyway it's a much cleaner mechanism.

peytondmurray (changes requested):

tfx/dependencies.py
smokestacklightnin: > Can just return this, I think. Also, doesn't `\n` need to be replaced with `' '`?

I'm not sure it makes sense to have a space `' '` appended to the end of every package name:

```python
In [1]: with open("requirements-docs.txt", "r") as fp:
   ...:     reqs = fp.readlines()
   ...: 

In [2]: reqs
Out[2]: 
['mkdocs\n',
 'mkdocstrings[python]\n',
 'mkdocs-material\n',
 'griffe-inherited-docstrings\n',
 'mkdocs-autorefs\n',
 'mkdocs-jupyter\n',
 'mkdocs-caption\n',
 'markdown-grid-tables\n']
```

I was trying to dynamically reproduce what was there before:

```python
def make_extra_packages_docs() -> list[str]:
    """Get a list of packages required for building docs as HTML.

    Returns
    -------
    list[str]
        List of packages required for building docs
    """
    return [
        "mkdocs",
        "mkdocstrings[python]",
        "mkdocs-material",
        "griffe-inherited-docstrings",
        "mkdocs-autorefs",
        "mkdocs-jupyter",
        "mkdocs-caption",
        "pymdown-extensions",
        "markdown-grid-tables",
    ]
```

.github/workflows/cd-docs.yml
smokestacklightnin: Done

tfx/dependencies.py
peytondmurray: Whoops, I see that you're using `readlines()` - makes sense!

peytondmurray (approved):
Looks good, thanks!

.github/workflows/cd-docs.yml
```diff
@@ -39,7 +40,12 @@ jobs:
...
       - name: Install Dependencies
-        run: pip install mkdocs mkdocs-material mkdocstrings[python] griffe-inherited-docstrings mkdocs-autorefs mkdocs-jupyter mkdocs-caption markdown-grid-tables
+        run: pip install -r requirements-docs.txt
 
       - name: Deploy to GitHub Pages
         run: mkdocs gh-deploy --force
+        if: (github.event_name != 'pull_request')
+
+      - name: Build docs to check for errors
+        run: mkdocs build --verbose
```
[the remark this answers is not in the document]
smokestacklightnin: Removed.

.github/workflows/cd-docs.yml
```diff
@@ -39,7 +40,12 @@ jobs:
             mkdocs-material-
 
       - name: Install Dependencies
-        run: pip install mkdocs mkdocs-material mkdocstrings[python] griffe-inherited-docstrings mkdocs-autorefs mkdocs-jupyter mkdocs-caption markdown-grid-tables
+        run: pip install -r requirements-docs.txt
 
       - name: Deploy to GitHub Pages
         run: mkdocs gh-deploy --force
+        if: (github.event_name != 'pull_request')
```
peytondmurray, line 47: This deploys the docs on every push to any branch.

peytondmurray (changes requested):
I think this deploys the docs on every push to any branch. We should only publish when there's a push to `master`.

peytondmurray (approved):
Looks great. Thank you!

lego0901 (approved):
Thanks!

PR #6924, opened 2024-10-02
Deprecate KFP v1 SDK support

nikelite:
New TFX version discontinues KFP v1 SDK support.


tfx/experimental/templates/container_based_test_case.py
briron, line 1: It is probably not related to this PR, but the test `tests (3.10, not e2e, DEFAULT)` is failing in this PR because of KEYERROR, and `_BASE_CONTAINER_IMAGE = os.environ['KFP_E2E_BASE_CONTAINER_IMAGE']` seems to be a culprint of the failure.

briron (commented):
Thanks for the work! However, some tests failed or were canceled. Do you have any idea why?

tfx/experimental/templates/container_based_test_case.py
nikelite: I think we should update some CI settings and/or some test utility files. Since py_test collects non-e2e tests at runtime, the unit tests are complaining that there are no valid environments.

cc: @peytondmurray

peytondmurray: Tagging @smokestacklightnin for input

peytondmurray: Hmm, I think these environment variables for for telling the tests what external docker containers are running so that various external services can be called. I don't think we should be collecting these tests, as we are not currently running any docker containers during the testing CI jobs. Let me make a PR...

peytondmurray:
@nikelite Normally I'd make a PR against your feature branch, but you're merging your `master` into the upstream `master`, so I've just made a PR against your `master`: https://github.com/nikelite/tfx/pull/1

If you merge that, the tests should pass here. The exception will be with the `NIGHTLY` tests; our dependencies are in a weird state where their allowed versions are not precisely enough defined for `pip`'s dependency solver to quickly find a solution, nor are they loosely defined enough for a solution to be easily found. As a result the `NIGHTLY` tests spend a huge amount of time downloading tons of versions of the various dependencies to try to find a solution. This is something we should make an issue for because currently it blocks folks from using `NIGHTLY`.

PR #6926, opened 2024-10-09
Update dependencies for the nightly build

nikelite:
This PR updates dependencies for the nightly build because there are too many nightly versions, and pip packages need to backtrack.


nikelite:
Technically, the NIGHT build should use TF 2.16.0.dev, as other recent nightly libraries are using it. However, TFMA is not yet compatible with this version, so we cannot enable nightly tests. I will revisit this once TFMA support is ready.

roseayeon (approved):

PR #6928, opened 2024-10-10
Update Docker files to support new toml project files

nikelite:
This PR updates Dockerfiles to support new TOML project files to enable integrated tests with Vertex ML pipelines.


nikelite:
@roseayeon , could you approve this PR?

This PR is for fixing Vertex end-to-end test that hasn't been included in the github action CI. 

I manually verified that this PR works with Vertex ML pipelines.

https://pantheon.corp.google.com/vertex-ai/pipelines/locations/us-central1/runs/taxi-template-vertex-e2e-1728959917-l627i0vs7g-20241015023945?project=tfx-testing

roseayeon (approved):

tfx/tools/docker/requirements.txt
```diff
@@ -0,0 +1,356 @@
+# The following versions are manually captured to avoid unfinished dependency resolution.
```
peytondmurray, line 1: It might be good to note here that this file is here because our project dependencies are complex, and without using these constraints, pip fails to solve when installing the project; otherwise new contributors may think these are actual project dependencies.

peytondmurray:
Two comments:

1. So this adds a constraints file so that we can run tests related to vertex pipelines. Do we have any tests for vertex pipelines?
2. There are a bunch of pinned dependencies in the constraints file - are they all necessary? For example, I see `pandas`, `h5py`, `markdown`, etc is in there...

peytondmurray (approved):
Looks good, see other comments though.

tfx/tools/docker/requirements.txt
nikelite: Thanks! I've added comments.

PR #6930, opened 2024-10-16
Add conftest.py to avoid absl.flags._exceptions.UnparsedFlagAccessError

nikelite:
- Add conftest.py to avoid `absl.flags._exceptions.UnparsedFlagAccessError`
- Re-enable related unit-tests


peytondmurray (commented):
These changes don't look like they modify the pytest configuration - did I miss something here? Only looks like install constraints were changed.

nikelite:
> These changes don't look like they modify the pytest configuration - did I miss something here? Only looks like install constraints were changed.

Sorry for the confusion. I've added a conftest.py file that prevents the `UnparsedFlagAccessError` from being raised during pytest execution, and I've updated the PR title to reflect this change.

tfx/conftest.py
```diff
@@ -0,0 +1,8 @@
+"""Test configuration."""
+from absl import flags
+
+def pytest_configure(config):
+  del config
```
peytondmurray, line 5: This deletes the pytest `Config` object - how come this is needed here? Won't this destroy all pytest and pytest plugin configuration?

nikelite: I initially thought the config object was only used within the function, but to prevent potential issues, I've removed it. Thanks!

PR #6936, opened 2024-10-21
Fix: Ensure proper cleanup of annotation artifact class for pytest multi-test contexts

nikelite:
This PR resolves an issue with the Invalid annotation artifact class by adding explicit tearDown steps. This is necessary for compatibility with pytest, which can execute multiple tests within a single context.


tfx/types/artifact_test.py
```diff
@@ -164,6 +157,10 @@ def decode(self, value: bytes):
 
 class ArtifactTest(tf.test.TestCase):
 
+  def tearDown(self):
+    # This cleans up __subclasses__() that has InvalidAnnotation artifact classes.
+    gc.collect()
+
```
smokestacklightnin, line 163: This appears to have worked since the tests pass. 

In the future, we can look at using [pytest yield fixtures](https://docs.pytest.org/en/stable/how-to/fixtures.html#yield-fixtures-recommended) instead of setup and teardown.

smokestacklightnin (approved):

Issue #6920, opened 2024-09-30
Evaluator Module_File install_to_temp_directory Failure


adammkerr:
**System information**

- Have I specified the code to reproduce the issue (Yes, No): Yes
- Environment in which the code is executed (e.g., Local(Linux/MacOS/Windows),
  Interactive Notebook, Google Cloud, etc): Vertex AI Pipelines
- TensorFlow version: 2.15.1
- TFX Version: 1.15.1
- Python version: 3.10.14
- Python dependencies (from `pip freeze` output):

```
absl-py==1.4.0
annotated-types==0.7.0
anyio==4.4.0
apache-beam==2.58.1
argon2-cffi==23.1.0
argon2-cffi-bindings==21.2.0
arrow==1.3.0
astunparse==1.6.3
async-lru==2.0.4
async-timeout==4.0.3
attrs==23.2.0
babel==2.16.0
backcall==0.2.0
beautifulsoup4==4.12.3
bleach==6.1.0
cachetools==5.5.0
certifi==2024.7.4
cffi==1.17.0
charset-normalizer==3.3.2
click==8.1.7
cloudpickle==2.2.1
colorama==0.4.6
comm==0.2.2
crcmod==1.7
debugpy==1.8.5
decorator==5.1.1
defusedxml==0.7.1
Deprecated==1.2.14
dill==0.3.1.1
dnspython==2.6.1
docker==4.4.4
docopt==0.6.2
docstring_parser==0.16
exceptiongroup==1.2.2
fastavro==1.9.5
fasteners==0.19
fastjsonschema==2.20.0
fire==0.6.0
flatbuffers==24.3.25
fqdn==1.5.1
gast==0.6.0
google-api-core==2.19.2
google-api-python-client==1.12.11
google-apitools==0.5.31
google-auth==2.34.0
google-auth-httplib2==0.2.0
google-auth-oauthlib==1.2.1
google-cloud-aiplatform==1.64.0
google-cloud-bigquery==3.25.0
google-cloud-bigquery-storage==2.25.0
google-cloud-bigtable==2.26.0
google-cloud-core==2.4.1
google-cloud-datastore==2.20.1
google-cloud-dlp==3.22.0
google-cloud-language==2.14.0
google-cloud-pubsub==2.23.0
google-cloud-pubsublite==1.11.1
google-cloud-recommendations-ai==0.10.12
google-cloud-resource-manager==1.12.5
google-cloud-spanner==3.48.0
[...]

PR #6937, opened 2024-10-22
Use 'local_pip_package_path' to install pip when the path refers to a remote file system

janasangeetha:
Updated [tfx/tfx/components/evaluator/executor.py](https://github.com/tensorflow/tfx/blob/c08360b3525a1d1af8d267933cc06f24af686dff/tfx/components/evaluator/executor.py#L123) to use local_pip_package_path to install pip when the path refers to a remote file system
#6920

### DIFF
--- a/tfx/components/evaluator/executor.py
+++ b/tfx/components/evaluator/executor.py
@@ -241,7 +241,7 @@ class Executor(base_beam_executor.BaseBeamExecutor):
     # may be created by the Beam multi-process DirectRunner) can find the
     # needed dependencies.
     # TODO(b/187122662): Move this to the ExecutorOperator or Launcher.
-    with udf_utils.TempPipInstallContext(extra_pip_packages):
+    with udf_utils.TempPipInstallContext(local_pip_package_path):
       with self._make_beam_pipeline() as pipeline:
         examples_list = []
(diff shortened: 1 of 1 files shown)


tfx/components/evaluator/executor.py
```diff
@@ -119,6 +119,7 @@ def Do(self, input_dict: Dict[str, List[types.Artifact]],
     # Make sure user packages get propagated to the remote Beam worker.
     unused_module_path, extra_pip_packages = udf_utils.decode_user_module_key(
         exec_properties.get(standard_component_specs.MODULE_PATH_KEY, None))
+    local_pip_package_path = []
```
nikelite, line 122: io_utils.ensure_local() returns string value, so you may want to additional list instead of using `local_pip_package_path`

tfx/components/evaluator/executor.py
```diff
@@ -119,6 +119,7 @@ def Do(self, input_dict: Dict[str, List[types.Artifact]],
...
     for pip_package_path in extra_pip_packages:
       local_pip_package_path = io_utils.ensure_local(pip_package_path)
```
nikelite, line 124: This path should be added to the list.

nikelite (approved):

PR #6940, opened 2024-10-26
Fix xFail errors from pytest 

nikelite:
* Re-enable TensorFlow eager mode to prevent unintended failures in other tests.
* Update the test class module name to align with the tests
* Include test data files in the TFX package to prevent test failures
* Resolve artifact type name conflicts from pytest.


.github/workflows/ci-test.yml
```diff
@@ -62,4 +62,4 @@ jobs:
     - name: Run unit tests
       shell: bash
       run: |
-        pytest -m "${{ matrix.which-tests }}"
+        pytest -vv -m "${{ matrix.which-tests }}" tfx
```
smokestacklightnin, line 65: We had specifically not chosen to put verbose output originally because there often could be an overwhelming amount of output on GitHub and developers can run pytest with verbose output on their local machines. If you would really like verbose output, leave it, by all means.

CC: @peytondmurray

tfx/experimental/distributed_inference/graphdef_experiments/subgraph_partitioning/beam_pipeline_test.py
```diff
@@ -25,6 +25,7 @@
 from tfx.experimental.distributed_inference.graphdef_experiments.subgraph_partitioning import create_complex_graph
 from tfx.experimental.distributed_inference.graphdef_experiments.subgraph_partitioning import graph_partition
 
+tf.compat.v1.enable_eager_execution()  # Re-enable eager mode
```
smokestacklightnin, line 28: If we need eager execution mode for _some_ tests, my strong preference and suggestion is to use a [pytest yield fixture](https://docs.pytest.org/en/stable/how-to/fixtures.html#yield-fixtures-recommended) as a decorator to enable and disable it for the tests that need it. The fixture can be stored in a `conftest.py` and reused anywhere it is needed.

CC: @peytondmurray

tfx/experimental/distributed_inference/graphdef_experiments/subgraph_partitioning/graph_partition_test.py
```diff
@@ -23,6 +23,7 @@
 from tfx.experimental.distributed_inference.graphdef_experiments.subgraph_partitioning import graph_partition
 from google.protobuf import text_format
 
+tf.compat.v1.enable_eager_execution()  # Re-enable eager mode
```
smokestacklightnin, line 26: Same pytest yield fixture comment as above.

smokestacklightnin (changes requested):

.github/workflows/ci-test.yml
smokestacklightnin: Also, is specifying the `tfx` directory strictly necessary here? I believe it is already specified in the [pytest config in `pyproject.toml`](https://github.com/tensorflow/tfx/blob/ee6eaf2c81916170571cc9288112c8a43504c3ef/pyproject.toml#L41)

tfx/experimental/distributed_inference/graphdef_experiments/subgraph_partitioning/beam_pipeline_test.py
peytondmurray: :point_up: To add context to this point:

- `tfx` has a hard dependency on `tensorflow>=2.15, <2.16`. Eager execution is [enabled by default](https://www.tensorflow.org/api_docs/python/tf/compat/v1/enable_eager_execution) for `tensorflow>=2`, so I don't think this is necessary.
- This setting unilaterally applies global changes to the testing environment, impacting much more than just the tests here. Even if `tfx` did allow users to install `tensorflow<2`, I'd still vote to move this into a test fixture to limit the impact on all the other tests in the test run.

On a related note, how many folks are still running TF<2? TF 2.0 came out over 5 years ago - if `tfx` still has tests that target deprecated 1.0 functionality, it may be worth starting an issue/discussion to remove them, as I don't think `tfx` can't be expected to run these at this point.

nikelite: The issue stemmed from the create_complex_graph module, which forcibly disables eager execution upon initialization. I've added a pytest yield fixture to re-enable it when the tests using this module end.

https://github.com/tensorflow/tfx/blob/ee6eaf2c81916170571cc9288112c8a43504c3ef/tfx/experimental/distributed_inference/graphdef_experiments/subgraph_partitioning/create_complex_graph.py#L25

tfx/experimental/distributed_inference/graphdef_experiments/subgraph_partitioning/beam_pipeline_test.py
```diff
@@ -25,7 +25,11 @@
 from tfx.experimental.distributed_inference.graphdef_experiments.subgraph_partitioning import create_complex_graph
 from tfx.experimental.distributed_inference.graphdef_experiments.subgraph_partitioning import graph_partition
 
-tf.compat.v1.enable_eager_execution()  # Re-enable eager mode
+@pytest.fixture
+def enable_eager_execution_at_test_end():
+  yield
+  # Re-enable eager mode that create_complex_graph diabled.
+  tf.compat.v1.enable_eager_execution()
```
smokestacklightnin, line 33: I think it's preferable for this to be located in [`conftest.py`](https://github.com/tensorflow/tfx/blob/master/conftest.py) so that it needs to only appear once in the code.

smokestacklightnin: It also appears to me that you've defined the fixture, but haven't told pytest to use it. You can do so by either the traditional method of [requesting (adding as a method argument)](https://docs.pytest.org/en/stable/how-to/fixtures.html#quick-example), or with a [`usefixtures` mark](https://docs.pytest.org/en/stable/how-to/fixtures.html#use-fixtures-in-classes-and-modules-with-usefixtures).

tfx/experimental/distributed_inference/graphdef_experiments/subgraph_partitioning/graph_partition_test.py
```diff
@@ -23,7 +23,12 @@
 from tfx.experimental.distributed_inference.graphdef_experiments.subgraph_partitioning import graph_partition
 from google.protobuf import text_format
 
-tf.compat.v1.enable_eager_execution()  # Re-enable eager mode
+@pytest.fixture
+def enable_eager_execution_at_test_end():
+  yield
+  # Re-enable eager mode that create_complex_graph diabled.
+  tf.compat.v1.enable_eager_execution()
+
```
smokestacklightnin, line 32: Same comments as above regarding the fixture

smokestacklightnin (changes requested):

tfx/experimental/distributed_inference/graphdef_experiments/subgraph_partitioning/beam_pipeline_test.py
nikelite: `[pytest yield fixture](https://docs.pytest.org/en/stable/how-to/fixtures.html#yield-fixtures-recommended)` is not working, since disable_eager_execution() isn't working as expected. This might be due to disable_eager_execution() being called during the collection phase of pytest, as it's also called during initialization. This code is from an older part of the project, and I'm unsure if it's safe to remove. However, since it's within experimental code, I'll comment it out for now.

.github/workflows/ci-test.yml
nikelite: I've reverted this change.

tfx/experimental/distributed_inference/graphdef_experiments/subgraph_partitioning/beam_pipeline_test.py
nikelite: I discovered that enable_eager_execution() is not compatible with this approach. All tests failed with the following error message: 'ValueError: tf.enable_eager_execution must be called at program startup.' 

Since tests under experimental folders passed without the enable_eager_execution() call, I just cleaned up the pytest.fixture code.

Issue #6943, opened 2024-10-27
Abstract base classes have incorrectly decorated methods


smokestacklightnin:
If the bug is related to a specific library below, please raise an issue in the
respective repo directly: TFX

**System information**

- Have I specified the code to reproduce the issue (Yes, No): Yes
- Environment in which the code is executed (e.g., Local(Linux/MacOS/Windows),
  Interactive Notebook, Google Cloud, etc): All
- TensorFlow version: All
- TFX Version: 1.15.1
- Python version: All
- Python dependencies (from `pip freeze` output): All

**Describe the current behavior**

There are currently violations of Ruff rules [B024](https://docs.astral.sh/ruff/rules/abstract-base-class-without-abstract-method/) and [B027](https://docs.astral.sh/ruff/rules/empty-method-without-abstract-decorator/), which have to do with incorrectly decorated class methods/properties. Fixing them currently causes errors (See #6942), so that issue should be dealt with before this one. 

**Describe the expected behavior**

See [python docs](https://docs.python.org/3/library/abc.html#abc.abstractproperty) for the correct way to make methods and properties abstract.

**Other info / logs**

The code violating B024 and B027 is listed below:

```
tfx/orchestration/portable/base_executor_operator.py:87:3: B027 `BaseExecutorOperator.handle_stop` is an empty method in an abstract base class, but has no abstract decorator
   |
85 |       return self
86 |   
87 |     def handle_stop(self) -> None:
   |  ___^
88 | |     """Executor Operator specific logic to clean up after it is stopped."""
89 | |     pass
   | |________^ B027
   |

tfx/tools/cli/handler/dag_runner_patcher.py:59:3: B027 `DagRunnerPatcher._before_run` is an empty method in an abstract base class, but has no abstract decorator
   |
57 |       self._call_real_run = call_real_run
58 |   
59 |     def _before_run(self, runner: tfx_runner.TfxRunner,
   |  ___^
60 | |                   pipeline: Union[pipeline_pb2.Pipeline, tfx_pipeline.Pipeline],
61 | |                   context: MutableMapping[str, Any]) -> None:
62 | |     pass
   | |________^ B027
63 |   
64 |     def _after_run(self, runner: tfx_runner.TfxRunner,
   |

tfx/tools/cli/handler/dag_runner_patcher.py:64:3: B027 `DagRunnerPatcher._after_run` is an empty method in an abstract base class, but has no abstract decorator
   |
62 |       pass
63 |   
[...]

Issue #6942, opened 2024-10-27
Code inistantiates abstract base classes


smokestacklightnin:
If the bug is related to a specific library below, please raise an issue in the
respective repo directly: TFX

**System information**

- Have I specified the code to reproduce the issue (Yes, No): Yes
- Environment in which the code is executed (e.g., Local(Linux/MacOS/Windows),
  Interactive Notebook, Google Cloud, etc): All
- TensorFlow version: All
- TFX Version: 1.15.1
- Python version: All
- Python dependencies (from `pip freeze` output): Any

**Describe the current behavior**

Several tests (listed below) instantiate abstract base classes, which shouldn't happen. 

I tried fixing fixing Ruff rules [B024](https://docs.astral.sh/ruff/rules/abstract-base-class-without-abstract-method/) and [B027](https://docs.astral.sh/ruff/rules/empty-method-without-abstract-decorator/) (see #6943), but because there are instantiated ABCs, fixing those rules caused errors.

**Describe the expected behavior**

Only concrete classes should be instantiated.

**More info**

The tests and/or modules that call code that instantiate ABCs are:
```
tfx/dsl/component/experimental/decorators_test.py::ComponentDecoratorTest::testBeamComponentBeamExecutionSuccess
tfx/dsl/component/experimental/decorators_test.py::ComponentDecoratorTest::testBeamExecutionBeamComponentWithInputArtifactAndParameters
tfx/dsl/component/experimental/decorators_test.py::ComponentDecoratorTest::testBeamExecutionNonNullableReturnError
tfx/dsl/component/experimental/decorators_test.py::ComponentDecoratorTest::testBeamExecutionSuccess
tfx/dsl/component/experimental/decorators_test.py::ComponentDecoratorTest::testComponentAnnotation
tfx/dsl/component/experimental/decorators_test.py::ComponentDecoratorTest::testJsonCompatParameter
tfx/dsl/component/experimental/decorators_test.py::ComponentDecoratorTest::testJsonCompatible
tfx/dsl/component/experimental/decorators_test.py::ComponentDecoratorTest::testListOfArtifacts
tfx/dsl/component/experimental/decorators_test.py::ComponentDecoratorTest::testOptionalInputsAndParameters
tfx/dsl/component/experimental/decorators_typeddict_test.py::ComponentDecoratorTest::testBeamComponentBeamExecutionSuccess
[...]

PR #6950, opened 2024-11-04
Fix Ruff rule B027 and B024

janasangeetha:
Fix Ruff rules
#6943

### DIFF
--- a/tfx/orchestration/portable/base_executor_operator.py
+++ b/tfx/orchestration/portable/base_executor_operator.py
@@ -84,6 +84,6 @@ class BaseExecutorOperator(abc.ABC):
     self._execution_watcher_address = execution_watcher_address
     return self
 
-  def handle_stop(self) -> None:# noqa: B027
+  def handle_stop(self) -> None:  # noqa: B027
     """Executor Operator specific logic to clean up after it is stopped."""
     pass
--- a/tfx/orchestration/portable/base_executor_operator.py
+++ b/tfx/orchestration/portable/base_executor_operator.py
@@ -22,7 +22,8 @@ from tfx.utils import abc_utils
 
 from google.protobuf import message
 
-class ParentBaseExecutorOperator(abc.ABC):
+
+class BaseExecutorOperator(abc.ABC):
   """The base class of all executor operators."""
 
   SUPPORTED_EXECUTOR_SPEC_TYPE = abc_utils.abstract_property()
@@ -83,28 +84,6 @@ class ParentBaseExecutorOperator(abc.ABC):
     self._execution_watcher_address = execution_watcher_address
     return self
 
-  @abc.abstractmethod
-  def handle_stop(self) -> None:
-    """Executor Operator specific logic to clean up after it is stopped."""
-    pass
-
-class BaseExecutorOperator(ParentBaseExecutorOperator):
-  """The child class for all abstract methods."""
-
-  def run_executor(
-      self,
-      execution_info: data_types.ExecutionInfo,
-  ) -> Optional[execution_result_pb2.ExecutorOutput]:
-    """Invokes the executor with inputs provided by the Launcher.
-
-    Args:
-      execution_info: A wrapper of the info needed by this execution.
-
-    Returns:
-      The output from executor.
-    """
-    pass
-
-  def handle_stop(self) -> None:
+  def handle_stop(self) -> None:# noqa: B027
     """Executor Operator specific logic to clean up after it is stopped."""
     pass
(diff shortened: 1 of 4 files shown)


tfx/orchestration/portable/base_executor_operator.py
```diff
@@ -84,6 +83,28 @@ def with_execution_watcher(
     self._execution_watcher_address = execution_watcher_address
     return self
 
+  @abc.abstractmethod
+  def handle_stop(self) -> None:
+    """Executor Operator specific logic to clean up after it is stopped."""
+    pass
+
+class BaseExecutorOperator(ParentBaseExecutorOperator):
```
lego0901, line 91: Is this really needed? I think it is sufficient to have `handle_stop` to be an abstractmethod, without needing to introduce `ParentBaseExecutorOperator`. I guess it is normal to have `handle_stop` implementation for any class that inherits `BaseExecutorOperator`.

tfx/tools/cli/handler/dag_runner_patcher.py
```diff
@@ -56,11 +56,13 @@ def __init__(self, call_real_run=True):
     self._run_called = False
     self._call_real_run = call_real_run
 
+  @abc.abstractmethod
   def _before_run(self, runner: tfx_runner.TfxRunner,
                   pipeline: Union[pipeline_pb2.Pipeline, tfx_pipeline.Pipeline],
                   context: MutableMapping[str, Any]) -> None:
     pass
 
+  @abc.abstractmethod
```
lego0901, line 65: Same here (and above `_before_run`). I believe all we need is `@abc.abstractmethod` decorators.

lego0901:
SG for B024. Please tackle me if anything is wrong with my comments on B027. Thanks!

janasangeetha:
@lego0901 On adding `@abc.abstractmethod` decorators, we observe failure in test cases #6942.
We should define the abstract methods of an abstract class in a child class. Hence, the parent classes are introduced.

lego0901:
I see.. Thanks. Indeed those methods are intentionally no-ops unless they are overridden.
Is there any way to suppress those Ruff warnings similar to B024? They are clearly not abstract methods, but adding a unnecessary chain seems not good for maintenance.

tfx/types/system_artifacts.py
```diff
@@ -30,7 +30,7 @@ class SystemArtifact(abc.ABC):
   The subclasses, e.g, Dataset, Model, Statistics, e.t.c, match the MLMD types
   from third_party/ml_metadata/metadata_store/mlmd_types.py.
   """
-
+  # noqa: B024
```
lego0901, line 33: This wasn't effective. Please try it out with

```python
class SystemArtifact(abc.ABC):  # noqa: B024
```

tfx/types/system_executions.py
```diff
@@ -30,7 +30,7 @@ class SystemExecution(abc.ABC):
   The subclasses, e.g, Train, Transform, Process, e.t.c, match the MLMD types
   from third_party/ml_metadata/metadata_store/mlmd_types.py.
   """
-
+  # noqa: B024
```
lego0901, line 33: ditto

tfx/orchestration/portable/base_executor_operator.py
```diff
@@ -84,6 +83,28 @@ def with_execution_watcher(
...
+
+    Args:
+      execution_info: A wrapper of the info needed by this execution.
+
+    Returns:
+      The output from executor.
+    """
+    pass
+
   def handle_stop(self) -> None:
```
lego0901, line 108: You can suppress B027 check via

```python
  def handle_stop(self) -> None:  # noqa: B027
```

tfx/tools/cli/handler/dag_runner_patcher.py
lego0901: ditto

lego0901:
Please make sure that everything is clear if you run
```
ruff check tfx --exclude tfx/examples/airflow_workshop/taxi/notebooks/*.ipynb --select "B024,B027"
```

Thanks!

tfx/orchestration/portable/base_executor_operator.py
```diff
@@ -84,6 +84,6 @@ def with_execution_watcher(
     self._execution_watcher_address = execution_watcher_address
     return self
 
-  def handle_stop(self) -> None:
+  def handle_stop(self) -> None:# noqa: B027
```
lego0901, line 87: And also a nit comment; as a coding convention, we were using two spaces before any comment `#`. This was warned when we are on the internal Google land, but I am not sure if it is captured by the linter here.

@smokestacklightnin could you let me know if this comment style is acceptable by the linter? `pylint --rcfile ./pylintrc tfx/types/system_executions.py` didn't warn me

lego0901 (approved):
Thanks!

tfx/orchestration/portable/base_executor_operator.py
smokestacklightnin: I don't believe the number of spaces matters. I tried with zero, one, and two spaces, and all three options passed `ruff check`.

PR #6952, opened 2024-11-06
Fix Ruff B008 errors

janasangeetha:
#6945

### DIFF
--- a/tfx/dsl/component/experimental/decorators_test.py
+++ b/tfx/dsl/component/experimental/decorators_test.py
@@ -140,8 +140,10 @@ def verify_beam_pipeline_arg(a: int) -> OutputDict(b=float):  # pytype: disable=
 
 def verify_beam_pipeline_arg_non_none_default_value(
     a: int,
-    beam_pipeline: BeamComponentParameter[beam.Pipeline] = beam.Pipeline(),
+    beam_pipeline: BeamComponentParameter[beam.Pipeline] = 0,
 ) -> OutputDict(b=float):  # pytype: disable=invalid-annotation,wrong-arg-types
+  if beam_pipeline == 0:
+    beam_pipeline = beam.Pipeline()
   del beam_pipeline
   return {'b': float(a)}
 
--- a/tfx/dsl/component/experimental/decorators_typeddict_test.py
+++ b/tfx/dsl/component/experimental/decorators_typeddict_test.py
@@ -140,8 +140,10 @@ def verify_beam_pipeline_arg(a: int) -> TypedDict('Output6', dict(b=float)):  # 
 
 def verify_beam_pipeline_arg_non_none_default_value(
     a: int,
-    beam_pipeline: BeamComponentParameter[beam.Pipeline] = beam.Pipeline(),
+    beam_pipeline: BeamComponentParameter[beam.Pipeline] = 0,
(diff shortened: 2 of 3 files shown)


tfx/dsl/component/experimental/decorators_test.py
```diff
@@ -140,8 +140,10 @@ def verify_beam_pipeline_arg(a: int) -> OutputDict(b=float):  # pytype: disable=
 
 def verify_beam_pipeline_arg_non_none_default_value(
     a: int,
-    beam_pipeline: BeamComponentParameter[beam.Pipeline] = beam.Pipeline(),
+    beam_pipeline: BeamComponentParameter[beam.Pipeline] = 0,
```
lego0901, line 143: `beam_pipeline: BeamComponentParameter[beam.Pipeline] = None` looks better

tfx/dsl/component/experimental/decorators_typeddict_test.py
```diff
@@ -140,8 +140,10 @@ def verify_beam_pipeline_arg(a: int) -> TypedDict('Output6', dict(b=float)):  #
 
 def verify_beam_pipeline_arg_non_none_default_value(
     a: int,
-    beam_pipeline: BeamComponentParameter[beam.Pipeline] = beam.Pipeline(),
+    beam_pipeline: BeamComponentParameter[beam.Pipeline] = 0,
```
lego0901, line 143: ditto

lego0901 (approved):
Thanks! Just added some minor comments

tfx/dsl/component/experimental/decorators_typeddict_test.py
janasangeetha: Error:
`
FAILED tfx/dsl/component/experimental/decorators_typeddict_test.py::ComponentDecoratorTest::testBeamPipelineDefaultIsNotNoneFails - TypeError: Argument a should be a Channel of type <class 'tfx.types.standard_artifacts.Integer'> (got 1)
`

tfx/dsl/component/experimental/decorators_test.py
lego0901: Oh I see. The typing error is pretty clear. You cannot assign an integer to other type.

If it is to avoid B008, then how's about doing it in this way?

```python
_TestBeamPipelineArgs = ['--my_testing_beam_pipeline_args=foo']
_TestBeamEmptyPipeline = beam.Pipeline()  # <- Newly added

...

def verify_beam_pipeline_arg_non_none_default_value(
    a: int,
    beam_pipeline: BeamComponentParameter[beam.Pipeline] = _TestBeamEmptyPipeline,
) -> OutputDict(b=float):  # pytype: disable=invalid-annotation,wrong-arg-types
  del beam_pipeline
  return {'b': float(a)}
```

tfx/dsl/component/experimental/decorators_typeddict_test.py
lego0901: same here

tfx/dsl/component/experimental/decorators_test.py
janasangeetha: Thanks! That works. I have updated.

tfx/dsl/component/experimental/decorators_typeddict_test.py
janasangeetha: Updated

PR #7726, opened 2024-12-03
TFX 1.16.0 Release

vkarampudi:


nikelite (approved):
Let's merge the following changes first and then release the package. Thanks!

https://github.com/tensorflow/tfx/pull/7727

RELEASE.md
```diff
@@ -41,8 +41,7 @@
 
 ## Deprecations
 
-
-*   KFP v1 SDK 
+*   KFP v1 SDK
```
nikelite, line 44: To provide further context, let's include:

* KubeflowDagRunner (KFP v1 SDK) is deprecated. Use KubeflowV2DagRunner (KFP v2 pipeline spec) instead.

vkarampudi: sure @nikelite, Updated

RELEASE.md
```diff
@@ -41,8 +41,7 @@
...
 *   TensorFlow Estimator is Deprecated
```
nikelite, line 45: To provide further context, let's include:

Since Estimators will no longer be available in TensorFlow 2.16 and later versions, we have deprecated examples and templates that use them. We encourage you to explore Keras as a more modern and flexible high-level API for building and training models in TensorFlow.

Issue #6909, opened 2024-08-29
Filter/Data Quality Component


pritamdodeja:
If the feature is related to a specific library below, please raise an issue in
the respective repo directly:

[TensorFlow Data Validation Repo](https://github.com/tensorflow/data-validation/issues)

[TensorFlow Model Analysis Repo](https://github.com/tensorflow/model-analysis/issues)

[TensorFlow Transform Repo](https://github.com/tensorflow/transform/issues)

[TensorFlow Serving Repo](https://github.com/tensorflow/serving/issues)

**System information**

- TFX Version (you are using): 1.15.1
- Environment in which you plan to use the feature (e.g., Local
  (Linux/MacOS/Windows), Interactive Notebook, Google Cloud, etc..):
Local, GCP

- Are you willing to contribute it (Yes/No): Yes

**Describe the feature and the current behavior/state.**
Whenever we have a tabular dataset, we compute statistics, and then a schema, and then we can validate examples.  We can fine-tune the schema in order to be integrate domain knowledge into the validation process.  Why cannot we filter "bad" data in the pipeline itself?

**Will this change the current API? How?**
It would introduce a new component that would allow filtering based on schema.  The inputs into Transform and some of the other components that use examples could potentially change depending on user's preference.

**Who will benefit with this feature?**
Users who want to do the entire data processing under the management of ML Metadata.

**Do you have a workaround or are completely blocked by this?** :
A variety of workarounds exist, manually clean up the original data, write scripts in beam to filter out bad data, put it in a database and use SQL to do this.

**Name of your Organization (Optional)**
GCP Partner (Intuitive.cloud)

**Any Other info.**
Number of training steps etc., would be affected by filtering data.  However, that is also something that, in my opinion, should be more dynamic than it currently is in tfx.

janasangeetha:
Hi @pritamdodeja 
Thank you for showing interest and raising an issue. Currently, we are not accepting new implementation requests or feature additions. If you would like to contribute , please follow our [CONTRIBUTION GUIDELINES](https://github.com/tensorflow/tfx/blob/master/CONTRIBUTING.md) and submit a pull request.
Thank you again for your interest, and we look forward to seeing your contributions!

pritamdodeja:
[...]

PR #7868, opened 2026-06-23
[Experimental] Add FilterComponent for row-level example filtering

vkarampudi:
# Description
This PR introduces an experimental **`FilterComponent`** to TFX, addressing a long-standing feature request ([Issue #6909](https://github.com/tensorflow/tfx/issues/6909)) for row-level example filtering within pipelines.

Currently, TFX has no built-in component to filter anomalous or unwanted examples after ingestion. The `Transform` component (TFT) is strictly map-only and does not support dropping rows. As a result, users have had to write custom Apache Beam code or perform out-of-band pre-processing. The new `FilterComponent` provides a clean, native, and reusable way to filter datasets split-by-split using a user-defined Python function.

# Proposed Changes
1. **Component Definition (`component.py`):**
   - Created `FilterSpec` and `FilterComponent`.
   - Accepts an input `Examples` channel and a string import path to a Python filter function (`filter_fn_path`).
   - Outputs a `filtered_examples` `Examples` channel.
2. **Beam Executor (`executor.py`):**
   - Implements a high-performance Apache Beam pipeline.
   - Automatically decodes all splits present in the input `Examples` artifact (e.g., `train`, `eval`).
   - Dynamically imports the user-defined filter function and applies `beam.Filter`.
   - Writes the output splits using standard TFRecord Gzip compression while preserving all metadata (spans, versions, split names) for downstream components.
3. **Tests:**
   - `component_test.py`: Verifies component instantiation and spec parsing.
   - `executor_test.py`: Runs a full integration test with simulated `train` and `eval` splits, executing the Beam filter pipeline and verifying correct output records.

# Usage Example
```python
# 1. Define a filter function in a shared module (e.g., my_filters.py)
def age_filter_fn(serialized_example: bytes) -> bool:
  example = tf.train.Example()
  example.ParseFromString(serialized_example)
  age_feature = example.features.feature.get('age')
  if age_feature:
    return age_feature.int64_list.value[0] > 18
  return False

# 2. Add the FilterComponent to your pipeline
from tfx.components.experimental.filter.component import FilterComponent

filter_data = FilterComponent(
    examples=example_gen.outputs['examples'],
    filter_fn_path='my_filters.age_filter_fn'
)

# 3. Pass the filtered output to Transform/Trainer
transform = Transform(
    examples=filter_data.outputs['filtered_examples'],
    schema=schema_gen.outputs['schema'],
    preprocessing_fn=...
)
```

### DIFF
--- a/tfx/components/experimental/filter/__init__.py
+++ b/tfx/components/experimental/filter/__init__.py
@@ -1,0 +1,14 @@
+# Copyright 2026 Google LLC. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Filter component experimental module."""
--- a/tfx/components/experimental/filter/component.py
+++ b/tfx/components/experimental/filter/component.py
@@ -1,0 +1,83 @@
+# Copyright 2026 Google LLC. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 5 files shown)


rwitcher (approved):
LGTM

pritamdodeja:
Wow! Thank you for implementing this! Reviewed the code/approach and it is exactly what I imagined for the tf.train.Example case.  Do you think this component would also work with tf.train.SequenceExample also?  What I mean is that the filter criteria is still a non-sequence type, but the record itself can have sequences.  For example, imagine a data quality threshold of some sort. Another scenario which I remember implementing in a likely hacky way was polygon based filtering, where you want to crop data using regions in space.  Classic example would be filter data from the "ocean" in the taxicab dataset.  In any case, super happy to see this get merged!

